PR #2, opened 2015-08-24
Fix some typos / doc improvements

zoffixznet:


jhthorsen:
My documentation would be awful without you!

<3

PR #3, opened 2015-09-21
Relative files

mrenvoize:
Failing test showing mis-loading of relative files..


jhthorsen:
Very cool! Thank you. Makes it a lot more understandable.

jhthorsen:
I think 5d7d9db should resolve this. Please try it out and let me know what you think.

PR #7, opened 2015-09-28
Test for recursive validation

mrenvoize:
Add tests for validation of recursive schemata.  Proves the recursive validation works as expected, always good to have more tests to prevent regressions :)

### DIFF
--- a/t/validate-recursive.t
+++ b/t/validate-recursive.t
@@ -1,0 +1,53 @@
+use Mojo::Base -strict;
+use Test::Mojo;
+use Test::More;
+use JSON::Validator 'validate_json';
+
+{
+  use Mojolicious::Lite;
+  post '/' => sub {
+    my $c = shift;
+    my @errors = validate_json $c->req->json, 'data://main/spec.json';
+    $c->render(status => @errors ? 400 : 200, text => "@errors");
+  };
+}
+
+my $t = Test::Mojo->new;
+
+$t->post_ok('/', json => {})->status_is(400)->content_like(qr{/person});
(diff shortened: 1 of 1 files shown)


jhthorsen:
Please use .perltidyrc to format the code.

Other than that: :+1:

mrenvoize:
As requested.. I thought I already had :$

jhthorsen:
I don't think you have: I see four spaces in the code, but... https://github.com/jhthorsen/json-validator/blob/master/.perltidyrc#L6

Would be nice if you rebased this PR and force pushed, but that's not a big deal.

mrenvoize:
Oh, that's odd.. It looks like my pertidy isn't picking up the .perltidyrc file :(

Sorry about that, I'll re-base.

t/validate-recursive.t
```diff
@@ -0,0 +1,53 @@
...
+use Test::Mojo;
+use Test::More;
+use JSON::Validator 'validate_json';
+
+{
+  use Mojolicious::Lite;
+  post '/' => sub {
+    my $c = shift;
+    my @errors = validate_json $c->req->json, 'data://main/spec.json';
+    $c->render(status => @errors ? 400 : 200, text => "@errors");
```
jhthorsen: Isn't it better to do ```json => \@errors``` ? I believe that would contribute to better unit testing than content_like().

mrenvoize:
How about this, both setting json as the response type, and also checking the json error response more explicitly with json_is.

jhthorsen:
Okidoki! :+1:

PR #12, opened 2015-11-09
Ensure we catch perl object within arrays

mrenvoize:
Catch, I hope, the last outstanding case of perl objects failing json object validation


jhthorsen:
Looks great! Just need to install it locally and run through the swagger and convos tests before I merge it.

jhthorsen:
Can't find any issues. How strange of me to add that code in the first place :/

PR #13, opened 2015-12-16
Allow hash reference as arguments to coerce()

kberov:
as this is the only way to ask for only one type of coercion from
a mojo configuration file.
Keep backward compatibility.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -279,12 +279,10 @@ sub coerce {
 
   return $self->{coerce} ||= {} unless @args;
 
-  @args = (booleans => 1, numbers => 1, strings => 1) if $args[0] eq '1';    # back compat
-  while (@args) {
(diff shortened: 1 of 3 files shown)


kberov:
I added a second commit which proves that when using Mojo::JSON::MaybeXS we get the error "Expected boolean - got JSON::PP::Boolean.". Interesting enough, this happens only when the value is "true".
Not sure exactly how this happens as we know now Mojo::JSON uses JSON::PP for booleans too.
if I do not "use Mojo::JSON::MaybeXS" all is fine.
If @Grinnz can help, I will be happy.
Thank you all guys!

kberov:
The last commit actually fixes the issue.

jhthorsen:
I'm going to merge this because of the tests, but the regex is no good imo. I will try to make it a bit more robust based on @grinnz' feedback.

kberov:
Thanks!

PR #19, opened 2016-05-19
No need for _merge_error

mrenvoize:
I'm giving this a thorough testing now.


mrenvoize:
Took me a while to get to test this, but I think it's an improvement so am happy to merge

PR #20, opened 2016-07-12
_merge_errors syntax fix

mrenvoize:
The sprintf would never get called with a non-empty arrayref before this
fix


jhthorsen:
Thanks! Would be nice with a test as well so I don't mess up again. Just commit the test directly to master if you write one.

mrenvoize:
Haha, I totally agree about adding a test.. I'll do a bit more digging, but it wasn't obvious actually what case this ends up being called with an array of more than one error ;).  I certainly couldn't manage it with my massive schema..  I'm sure i'll track it down ;)

PR #21, opened 2016-07-12
Clarify 'anyOf' errors

mrenvoize:
This pull request goes some way towards clarifying anyOf error messages.. It's limits us down to the single error at the deepest failure point in the tree.

### DIFF
--- a/t/jv-anyof.t
+++ b/t/jv-anyof.t
@@ -80,8 +80,7 @@ my $schemaC = {
 
 @errors = $validator->validate("string not integer", $schemaC);
 
-is "@errors", "/: anyOf failed: Expected integer, got string.", "nesting: string not integer (or object)";
-# It would be nice to get 'object or integer' in the error above
+is "@errors", "/: anyOf failed: Expected object or integer, got string.", "nesting: string not integer (or object)";
 
 @errors = $validator->validate({id => 1, name => 'Bob'}, $schemaC);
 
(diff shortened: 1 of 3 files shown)


jhthorsen:
Sorry, I don't get why this change makes the error more clear. Maybe it will be more obvious if you provide a more complex test in addition to changing the existing tests.

mrenvoize:
Sure thing,.. I'll do a more in depth test shortly..

But.. in short.. for nested errors we go from:

{ path: '/', message: "anyOf failed: anyOf failed: anyOf failed: Expected String got Number" }

to:

{ path:'/a/b/c/d', message: "Expected String got Number" }

jhthorsen:
That sounds fantastic! Please prove it with a test, so we don't go back to the confusing message.

mrenvoize:
At last, I think I've got there..

It definitely needs a bit of golfing though.. the code is somewhat cringy in places but the functionality is working.

t/jv-anyof.t
```diff
@@ -80,7 +80,8 @@ my $schemaC = {
 
 @errors = $validator->validate("string not integer", $schemaC);
 
-is "@errors", "/: anyOf failed: Expected object or integer, got string.", "nesting: string not integer (or object)";
+is "@errors", "/: anyOf failed: Expected integer, got string.", "nesting: string not integer (or object)";
+# It would be nice to get 'object or integer' in the error above
```
jhthorsen: I think avoiding "object" makes it clearer.

mrenvoize: You could be right.. I dunno.. happy for you to make that call

mrenvoize: Want me to remove it?

jhthorsen:
Closing this because of inactivity.

Issue #15, opened 2016-01-20
Errors in anyOf validation. Do not understand the error messages


elowy01:
Hi,

It seems that in the validation process of a schema containing anyOf statements, the code issue errors for each of the attributes failing the test, regardless the anyOf branch that the test is.  For example in the script jv-anyof.t in your test suite. In one of the tests:

```
@errors = $validator->validate("too long", $schema);
```

The $validator generate the following message:

anyOf failed: ([0] String is too long: 8/5. [1] Expected number - got string.)'

and I am only interested in knowing that the string is too long and not in the additional 'Expected number -got string' error.

Is there a way of generating only the 'String is to long' error?

Thanks,

ernesto

jhthorsen:
That's an interesting question. I guess it could skip all the tests that doesn't match the input type. Not sure how simple it would be to implement it though.

I'm leaving this open, in case someone wants to give it a go.

elowy01:
Thanks for your reply, I will try to have it a go

Another question, 

Are the numbers in the error messages ([0],[1] and so on) related to the branches of the anyOf tests in a certain way?

Thanks,

e.

jhthorsen:
Yes, the number inside the brackets indicates which rule number that failed.

jhthorsen:
I started working on https://github.com/jhthorsen/json-validator/compare/anyof-oneof-allof-errors yesterday, but it's not my primary target to fix this, since I don't find it _that_ annoying.

If someone wants to complete the work, I would be happy to take a PR.

mrenvoize:
@elowy01 We've been working hard on this today to improve the situation.. everything is now released.. would appreciate your feedback.. I'm looking to close this issue soon.

mrenvoize:
Further work in https://github.com/jhthorsen/json-validator/tree/syntax and

mrenvoize:
Would appreciate your feedback on the rebased branch https://github.com/jhthorsen/json-validator/tree/all-any-one-of-rebase

It's looking promising to me so far.. though there's one outstanding failing test.. jv-object which I'm struggling to pin down.

I'll be running this against the openapi test suits too once this last test is passing again (any pointers would help there.. I'm happy to keep plodding, but you might spot it more easily than me).

I'm also a little bit confused by why a test in jv-oneof results in an anyof type error.. feels like a bug to me.. possibly a oneof error should be the result.

KES777:
I also do not understand what is happen. The error is:

```
[...]

PR #24, opened 2016-07-28
Prettier allOf, anyOf and oneOf error messages

jhthorsen:
See also #15 and #21 and https://github.com/jhthorsen/json-validator/commit/2148a9b53361c50f51ca83e47946e47094799fb2

It is very, very important to run all the swagger2 tests before releasing this change to CPAN. (2148a9b53361c50f51ca83e47946e47094799fb2 broke Swagger2, even if all the tests are ok)

Also, for now it only has prettier anyOf messages. (imo)


jhthorsen:
Closing this because of inactivity.

mrenvoize:
I'm looking at resurrecting this one again now, I can't replicate any of the problems we were having with ::Swagger2 in ::OpenAPI with this patchset and it definitely improved error messages.

PR #25, opened 2016-07-29
OpenAPI $ref fails directly under /paths and /definitions

taskula:
Currentyl openapi-ref.t is skipping references for not only our test spec, but also OpenAPI spec. We need to resolve $refs in OpenAPI spec in order to have correct rules, meanwhile not resolving $refs for our test spec.


taskula:
Link to original issue https://github.com/jhthorsen/swagger2/issues/89

jhthorsen:
Thanks! This was very helpful 👍

PR #33, opened 2016-11-03
Fix number coercion

melhesedek:
Probably needs more tests. Regex taken from [stackoverflow](http://stackoverflow.com/a/13340826).


jhthorsen:
This is perfect! Thank you 👍

PR #34, opened 2016-11-03
Make JSON::Validator::OpenAPI framework-independent

dboehmer:
Initially implements Mojolicious (based on existing code) and Dancer2.

### DIFF
--- a/lib/JSON/Validator/OpenAPI.pm
+++ b/lib/JSON/Validator/OpenAPI.pm
@@ -192,7 +192,7 @@ sub _validate_response_headers {
     }
     elsif ($input->{$name}) {
       push @errors, $self->validate($input->{$name}[0], $p);
-      $c->_set_response_data($c, 'header', $name => $input->{$name}[0] ? 'true' : 'false')
+      $self->_set_response_data($c, 'header', $name => $input->{$name}[0] ? 'true' : 'false')
         if $p->{type} eq 'boolean' and !@errors;
     }
   }
--- a/lib/JSON/Validator/OpenAPI.pm
+++ b/lib/JSON/Validator/OpenAPI.pm
@@ -30,7 +30,7 @@ sub validate_request {
       $exists = length $value if defined $value;
     }
     elsif ($in eq 'formData' and $type eq 'file') {
-      $value = $c->req->upload($name);
+      $value = $self->_extract_upload($c, $name);
       $exists = $value ? 1 : 0;
     }
     else {
@@ -91,44 +91,6 @@ sub validate_response {
   return @errors;
 }
 
-sub _extract_request_parameter {
-  my ($self, $c, $in) = @_;
-
-  return $c->req->url->query->to_hash  if $in eq 'query';
-  return $c->match->stack->[-1]        if $in eq 'path';
-  return $c->req->body_params->to_hash if $in eq 'formData';
-  return $c->req->headers->to_hash     if $in eq 'header';
-  return $c->req->json                 if $in eq 'body';
-  return {};
-}
-
-sub _set_request_parameter {
-  my ($self, $c, $p, $value) = @_;
-  my ($in, $name) = @$p{qw(in name)};
-
-  if ($in eq 'query') {
-    $c->req->url->query([$name => $value]);
-    $c->req->params->merge($name => $value);
-  }
-  elsif ($in eq 'path') {
-    $c->stash($name => $value);
-  }
-  elsif ($in eq 'formData') {
-    $c->req->params->merge($name => $value);
-    $c->req->body_params->merge($name => $value);
-  }
-  elsif ($in eq 'header') {
-    $c->req->headers->header($name => $value);
-  }
-  elsif ($in eq 'body') {
-    return;    # no need to write body back
-  }
-  else {
-    die
-      "Cannot set default for $in => $name. Please submit a ticket here: https://github.com/jhthorsen/mojolicious-plugin-openapi";
-  }
-}
-
 sub _validate_request_value {
   my ($self, $p, $name, $value) = @_;
   my $type = $p->{type} || 'object';
@@ -165,7 +127,7 @@ sub _validate_request_value {
 
 sub _validate_response_headers {
   my ($self, $c, $schema) = @_;
-  my $headers = $c->res->headers;
+  my $headers = $self->_extract_headers($c);
   my $input   = $headers->to_hash(1);
   my @errors;
 
--- a/lib/JSON/Validator/OpenAPI.pm
+++ b/lib/JSON/Validator/OpenAPI.pm
@@ -10,6 +10,28 @@ my %COLLECTION_RE = (pipes => qr{\|}, csv => qr{,}, ssv => qr{\s}, tsv => qr{\t}
 
 has _json_validator => sub { state $v = JSON::Validator->new; };
 
+{    # proxy methods for compatibility
+  my @proxy_methods = qw<
+    _get_request_uploads
+    _get_request_data
+    _set_request_data
+    _get_response_data
+    _set_response_data
+  >;
+
+  for my $method (@proxy_methods) {
+    no strict 'refs';
+    *{__PACKAGE__ . "::$method"} = sub {
+      warn "Using JSON::Validator::OpenAPI directly is DEPRECATED."
+        . " For the Mojolicious-specific methods use JSON::Validator::OpenAPI::Mojolicious";
+
+      shift @_;
+      require JSON::Validator::OpenAPI::Mojolicious;
+      JSON::Validator::OpenAPI::Mojolicious->$method(@_);
+    };
+  }
+}
+
 sub validate_input {
   my $self = shift;
   local $self->{validate_input} = 1;
--- a/lib/JSON/Validator/OpenAPI.pm
+++ b/lib/JSON/Validator/OpenAPI.pm
@@ -30,7 +30,7 @@ sub validate_request {
       $exists = length $value if defined $value;
     }
     elsif ($in eq 'formData' and $type eq 'file') {
-      $value = $self->_get_request_uploads($c, $name);
+      ($value) = $self->_get_request_uploads($c, $name);
       $exists = $value ? 1 : 0;
     }
     else {
--- a/lib/JSON/Validator/OpenAPI.pm
+++ b/lib/JSON/Validator/OpenAPI.pm
@@ -26,15 +26,15 @@ sub validate_request {
     my ($exists, $value);
 
     if ($in eq 'body') {
-      $value = $self->_extract_request_parameter($c, $in);
+      $value = $self->_get_request_data($c, $in);
       $exists = length $value if defined $value;
     }
     elsif ($in eq 'formData' and $type eq 'file') {
-      $value = $self->_extract_upload($c, $name);
+      $value = $self->_get_request_uploads($c, $name);
       $exists = $value ? 1 : 0;
     }
     else {
-      $value  = $cache{$in} ||= $self->_extract_request_parameter($c, $in);
+      $value  = $cache{$in} ||= $self->_get_request_data($c, $in);
       $exists = exists $value->{$name};
       $value  = $value->{$name};
     }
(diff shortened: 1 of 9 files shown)


lib/JSON/Validator/OpenAPI.pm
```diff
@@ -10,6 +11,28 @@ my %COLLECTION_RE = (pipes => qr{\|}, csv => qr{,}, ssv => qr{\s}, tsv => qr{\t}
...
+
+  for my $method (@proxy_methods) {
+    no strict 'refs';
+    *{__PACKAGE__ . "::$method"} = sub {
+      deprecated "Using JSON::Validator::OpenAPI directly is DEPRECATED."
+        . " For the Mojolicious-specific methods use JSON::Validator::OpenAPI::Mojolicious";
+
+      shift @_;
+      require JSON::Validator::OpenAPI::Mojolicious;
+      JSON::Validator::OpenAPI::Mojolicious->$method(@_);
```
jhthorsen: I don't get how this is supposed to work... doesn't those methods need $self at all? I think I would do this instead:

```
deprecated "...";
require JSON::Validator::OpenAPI::Mojolicious;
goto do { JSON::Validator::OpenAPI::Mojolicious->can($method) };
```

lib/JSON/Validator/OpenAPI.pm
```diff
@@ -10,6 +11,28 @@ my %COLLECTION_RE = (pipes => qr{\|}, csv => qr{,}, ssv => qr{\s}, tsv => qr{\t}
 
 has _json_validator => sub { state $v = JSON::Validator->new; };
 
+{    # proxy methods for compatibility
+  my @proxy_methods = qw<
```
jhthorsen: Minor thing, but could you use `qw()` for consistency?

lib/JSON/Validator/OpenAPI.pm
```diff
@@ -26,15 +49,15 @@ sub validate_request {
     my ($exists, $value);
 
     if ($in eq 'body') {
-      $value = $self->_extract_request_parameter($c, $in);
+      $value = $self->_get_request_data($c, $in);
       $exists = length $value if defined $value;
     }
     elsif ($in eq 'formData' and $type eq 'file') {
-      $value = $c->req->upload($name);
+      ($value) = $self->_get_request_uploads($c, $name);
```
jhthorsen: Why do you need `($value)` and not just `$value`?

lib/JSON/Validator/OpenAPI.pm
```diff
@@ -165,8 +150,7 @@ sub _validate_request_value {
 
 sub _validate_response_headers {
   my ($self, $c, $schema) = @_;
-  my $headers = $c->res->headers;
-  my $input   = $headers->to_hash(1);
+  my $input = $self->_get_response_data($c, 'header');
```
jhthorsen: 👍

lib/JSON/Validator/OpenAPI/Dancer2.pm
```diff
@@ -0,0 +1,72 @@
+package JSON::Validator::OpenAPI::Dancer2;
+
+use Carp qw(confess);
+use Hash::MultiValue;
```
jhthorsen: Did you miss a dependency? Or does Dancer2 already depend on this?

lib/JSON/Validator/OpenAPI/Dancer2.pm
```diff
@@ -0,0 +1,72 @@
...
+sub _get_request_data {
+  my ($self, $dsl, $in) = @_;
+
+  return $dsl->query_parameters->as_hashref_mixed if $in eq 'query';
+  return $dsl->route_parameters->as_hashref_mixed if $in eq 'path';
+  return $dsl->body_parameters->as_hashref_mixed  if $in eq 'formData';
+  return Hash::MultiValue->new($dsl->app->request->headers->flatten)->as_hashref_mixed
+    if $in eq 'header';
+  return $dsl->app->request->data if $in eq 'body';
+  confess "Unsupported \$in: $in. Please report at https://github.com/jhthorsen/json-validator";
```
jhthorsen: 👍

lib/JSON/Validator/OpenAPI/Dancer2.pm
```diff
@@ -0,0 +1,72 @@
...
+}
+
+sub _get_request_uploads {
+  my ($self, $dsl, $name) = @_;
+
+  return ($dsl->app->request->upload($name));    # context-aware
```
jhthorsen: I don't want anything to be context aware. Force array-ref or scalar.

lib/JSON/Validator/OpenAPI/Dancer2.pm
```diff
@@ -0,0 +1,72 @@
+package JSON::Validator::OpenAPI::Dancer2;
+
+use Carp qw(confess);
+use Hash::MultiValue;
+use Mojo::Base 'JSON::Validator::OpenAPI';
+
+### request ###
+
+sub _get_request_data {
+  my ($self, $dsl, $in) = @_;
```
jhthorsen: Just curious: What doe `$dsl` mean?

lib/JSON/Validator/OpenAPI/Mojolicious.pm
```diff
@@ -0,0 +1,67 @@
...
+  elsif ($in eq 'formData') {
+    $c->req->params->merge($name => $value);
+    $c->req->body_params->merge($name => $value);
+  }
+  elsif ($in eq 'header') {
+    $c->req->headers->header($name => $value);
+  }
+  elsif ($in eq 'body') { }    # no need to write body back
+  else {
+    die
```
jhthorsen: Carp::confess..? (I can fix this)

lib/JSON/Validator/OpenAPI/Mojolicious.pm
```diff
@@ -0,0 +1,67 @@
...
+
+  # TODO what else?
+}
+
+sub _set_response_data {
+  my ($self, $c, $in, $name => $value) = @_;
+
+  $in eq 'header' and $c->res->headers->header($name => ref $value ? @$value : $value);
+
+  # TODO what else?
```
jhthorsen: Good question... I wonder if we can make a helper method, instead of "sprinkling" Carp::confess all over. I can have a look at it, after the merge.

t/openapi-compatibility.t
```diff
@@ -0,0 +1,42 @@
...
+  "JSON::Validator::OpenAPI::Mojolicious->validate_request isn't deprecated"
+);
+
+sub module_calls_deprecated {
+  my $module = shift;
+
+  my $called = 0;
+  {
+    no warnings 'redefine';
+    *Mojo::Util::deprecated = sub { ++$called };
```
jhthorsen: Use `Mojo::Util::monkey_patch("Mojo::Util", deprecated => sub { ++$called })` instead.

t/openapi-formats.t
```diff
@@ -1,8 +1,8 @@
 use Mojo::Base -strict;
 use Test::More;
-use JSON::Validator::OpenAPI;
+use JSON::Validator::OpenAPI::Mojolicious;
 
-my $validator = JSON::Validator::OpenAPI->new;
+my $validator = JSON::Validator::OpenAPI::Mojolicious->new;
```
jhthorsen: Nice!

jhthorsen:
There's so many nice enhancements here @dboehmer! Thank you. Looking forward to completing this branch.

lib/JSON/Validator/OpenAPI/Dancer2.pm
dboehmer: `Dancer2` already depends on `Hash::MultiValue` but I'll add an explicit dependency anyway.

lib/JSON/Validator/OpenAPI.pm
dboehmer: That is what bugged my at the very beginning when I asked how to use JSON::Validator::OpenAPI. The calls to `validate_request` never use `$self`, I think. However the framework-specific methods only need `$c` passed.

I like that you suggest `do` because it's generally discouraged and few people seem to know this style of use. But as these methods work as simple class method I see no reason to change this.

lib/JSON/Validator/OpenAPI.pm
dboehmer: I actually like `qw<>` because `<>` does the same in Perl6 but for consistency's sake I tried to keep close to your style. You're right.

lib/JSON/Validator/OpenAPI/Dancer2.pm
dboehmer: That is an instance of [`Dancer2::Core::DSL`](https://metacpan.org/pod/Dancer2::Core::DSL) with the current request, etc. set in the background which provides all keywords from Dancer's *domain-specific language* (DSL). In this case it has the same purpose as `$c` in Mojolicious or Catalyst.

lib/JSON/Validator/OpenAPI/Dancer2.pm
dboehmer: Oh, the comment reminds that `upload()` is context-aware. I added the parens to enforce list context. See other comments about `($value) =  ...`

lib/JSON/Validator/OpenAPI/Mojolicious.pm
dboehmer: I'll do with the other changes.

lib/JSON/Validator/OpenAPI/Mojolicious.pm
dboehmer: Good idea.

lib/JSON/Validator/OpenAPI.pm
jhthorsen: Thank you.

lib/JSON/Validator/OpenAPI/Dancer2.pm
jhthorsen: Thank you.

t/openapi-compatibility.t
dboehmer: `Mojo::Util` is a treasure of wonderful tools. After you suggest using functions from this package multiple times I've now read the whole POD to avoid getting into this again;-)

lib/JSON/Validator/OpenAPI/Dancer2.pm
dboehmer: Settled in IRC: As it's already a dependency of `Dancer2` we won't depend on this ourself.

dboehmer:
I added a fix and merged your `master` branch into my PR branch. Can you please review again and merge if possible?

lib/JSON/Validator/OpenAPI/Mojolicious.pm
```diff
@@ -0,0 +1,72 @@
...
+
+### request ###
+
+sub _get_request_data {
+  my ($self, $c, $in) = @_;
+
+  return $c->req->url->query->to_hash(1)  if $in eq 'query';
+  return $c->match->stack->[-1]           if $in eq 'path';
+  return $c->req->body_params->to_hash(1) if $in eq 'formData';
+  return $c->req->headers->to_hash(1)     if $in eq 'header';
```
jhthorsen: Why are you passing in a true value to the to_hash() methods? The old code did not do this, and it breaks swagger2.pm header tests.

jhthorsen:
I've rebased the commits and made a new release with the code in this PR.

Thanks for the contribution @dboehmer 👍

lib/JSON/Validator/OpenAPI/Mojolicious.pm
dboehmer: I find it important to verify all occurrences of parameter keys. My suggested API shall return all the values as hashes with arrayrefs for multiple values. In Dancer2 I used `Hash::MultiValue->to_mixed_hashref`. `Mojo::Headers` implements a flag argument to produce "mixed hashrefs" in [`$headers->to_hash(1)`](https://metacpan.org/pod/Mojo::Headers#to_hash). I don't know how this is supposed to work for query or body params but I think we should have mixed hashrefs there, too.

jhthorsen: Please have a look at https://github.com/jhthorsen/json-validator/pull/38 @dboehmer.

Issue #36, opened 2016-11-22
Integer path param ending with \w warns and passes validation


ugexe:
(Note: I'm using `Mojolicious::Plugin::OpenAPI`, so it's possibly a bug from there)

I have an integer path parameter: `"parameters" : [ { "in": "path", "name": "id", "type": "integer" } ],`, and when I give it a value such as "42a" I get an unexpected warning and there is no validation failure.

I logged the value of $value on the lines before and after the line causing the warning and include it in the output below:

    perl -Ilocal/lib/perl5 -Ilib script/swagger.pl get /document/42a
    "42a" #L79 `p $value;`
    Argument "\x{34}\x{32}..." isn't numeric in addition (+) at local/lib/perl5/JSON/Validator/OpenAPI.pm line 80.
    42    #L81 `p $value;`

    # "\x{34}\x{32}" is 42

Seems like it might be an issue with coercing somewhere, but I couldn't figure out how to golf it

    perl -E 'my $x = "42a"; $x += 0; say $x'
    42

jhthorsen:
Fixed by using `Scalar::Util::looks_like_number()`

Thank you!

PR #37, opened 2016-11-22
Workaround bug with int params ending with chars

ugexe:
A path /document/{id} thats requires an integer could give a
warning and pass validation given a value such as "42a". This
switches the number validation routine used so it fails validation
for being a string.

https://github.com/jhthorsen/json-validator/issues/36

### DIFF
--- a/lib/JSON/Validator/OpenAPI.pm
+++ b/lib/JSON/Validator/OpenAPI.pm
@@ -71,7 +71,7 @@ sub validate_request {
     }
 
     if ($type and defined($value //= $p->{default})) {
-      if (($type eq 'integer' or $type eq 'number') and $value =~ /^-?\d/) {
+      if (($type eq 'integer' or $type eq 'number') and _is_number($value)) {
         $value += 0;
       }
       elsif ($type eq 'boolean') {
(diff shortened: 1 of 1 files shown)


ugexe:
Sorry, forgot to include the second commit earlier!

jhthorsen:
Thanks for your contribution! Could you also write a test for this?

I think copy/pasting these two lines that I link to below with your input above would be enough. Maybe add a `local $SIG{__WARN__} = sub { push @warn, $_[0] };` as well to make sure there's no warning detected.

https://github.com/jhthorsen/json-validator/blob/master/t/jv-integer.t#L38-L39

ugexe:
The problem is I can't golf it down. I can only get it to happen when using Mojolicious::Plugin::OpenAPI so I could only provide a test for that module:

    # https://github.com/jhthorsen/mojolicious-plugin-openapi/blob/master/t/path-parameters.t
    # The following test passes despite the integer constraint:
    $t->post_ok('/api/user/42a' => json => {})->status_is(200)->json_is('/id', '42a');

jhthorsen:
Closed in #36.

PR #40, opened 2017-01-13
Fix for infinite recursion

mrenvoize:
I found that if you had two ref's pointing to the same location from the same level in the json structure then you could end up in an infinite recursion.

I couldn't work out how to create a test for this :(, perl doesn't detect the recursion inside this while loop, but you can see it in a console if you put the right print statements in.

This pull request just adds what I believe to be the fix, I'd appreciate your expertise in helping write a test for this.


mrenvoize:
Grr.. let me get to the bottom of those failing tests first :(

mrenvoize:
Grr, let me get to the bottom of those failing tests first! :(

jhthorsen:
You need to submit a failing test as well.

PR #43, opened 2017-02-09
resolve #42 - honor Cache-Control headers ...

### LINKED ISSUE #42: Cached specifications never expire
Probably makes sense to have some sort of sensible default expiry time on cached specifications, otherwise you have to make sure you clear up cached specifications manually before calling JSON::Validator or risk getting an outdated spec for the given namespace.

leejo:
... when remotely fetching specs. We honor the no-cache, no-store,
private, and revalidate controls by not caching, and honor max-age
by setting another file to store the expiry time of the cached file

add a test (requires TEST_ONLINE=1) to check for this logic

remove t/spec/invalid-ref.t as is an empty test with no other refs
in the repo or git log, and causes prove to fail

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -18,7 +18,7 @@ use constant VALIDATE_IP       => eval 'require Data::Validate::IP;1';
 
 use constant DEBUG => $ENV{JSON_VALIDATOR_DEBUG} || 0;
 
-our $VERSION = '0.92';
+our $VERSION   = '0.92';
 our @EXPORT_OK = 'validate_json';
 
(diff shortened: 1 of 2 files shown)


leejo:
> I don't think this is the right solution. I don't want my application to change behaviour because some external file changed.

If you don't want that to happen then you *shouldn't* be requesting specifications from a remote source, i would have thought this is pretty obvious. You request information over http, it can and will change.

jhthorsen:
I'm closing this in favour of #44 or #45.

jhthorsen:
Side note: I think this is a nice change for the problem it tries to tackle. I just don't think it's the right problem.

PR #44, opened 2017-02-09
resolve #42 - default cache_path should not be writeable

### LINKED ISSUE #42: Cached specifications never expire

leejo:
we should not store remotely cached specs in the default cache dir
as this can be confusing to end users as we don't honor the headers
for Cache-Control

make sure we skip the default cache dir when caching files, add a
test to ensure this doesn't happen and that if the cache_paths is
overriden we *do* cache remote files

update documentation to clarify the way cache_paths works w/r/t to
local and remote specs

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -18,9 +18,10 @@ use constant VALIDATE_IP       => eval 'require Data::Validate::IP;1';
 
 use constant DEBUG => $ENV{JSON_VALIDATOR_DEBUG} || 0;
 
-our $VERSION = '0.92';
+our $VERSION   = '0.92';
 our @EXPORT_OK = 'validate_json';
 
+my $DEFAULT_CACHE_DIR = File::Spec->catdir(File::Basename::dirname(__FILE__), qw(Validator cache));
 my $HTTP_SCHEME_RE = qr{^https?:};
 
 sub E { JSON::Validator::Error->new(@_) }
@@ -35,7 +36,7 @@ has cache_paths => sub {
   my $self = shift;
   my @paths = split /:/, ($ENV{JSON_VALIDATOR_CACHE_DIR} || '');
   push @paths, $self->{cache_dir} if $self->{cache_dir};
-  push @paths, File::Spec->catdir(File::Basename::dirname(__FILE__), qw(Validator cache));
+  push @paths, $DEFAULT_CACHE_DIR;
   return \@paths;
 };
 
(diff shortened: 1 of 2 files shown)


lib/JSON/Validator.pm
```diff
@@ -180,9 +180,12 @@ sub _load_schema_from_url {
...
   die $tx->error->{message} if $tx->error;
 
-  if ($cache_dir and -w $cache_dir) {
-    $cache_file = File::Spec->catfile($cache_dir, $cache_file);
-    Mojo::Util::spurt($tx->res->body, $cache_file);
+  for my $cache_dir (@{$self->cache_paths}) {
+    if ($cache_dir and -w $cache_dir and $cache_dir ne $DEFAULT_CACHE_DIR) {
+      $cache_file = File::Spec->catfile($cache_dir, $cache_file);
+      Mojo::Util::spurt($tx->res->body, $cache_file);
+      last;
```
leejo: Not 100% sure about the last here - do we want to cache specs in *all* cache dirs?

jhthorsen: No, I don't think so.

lib/JSON/Validator.pm
```diff
@@ -980,8 +983,9 @@ To download a file and add it to the cache, do this:
 
   $ curl http://swagger.io/v2/schema.json > /cache/dir/$(md5 -qs http://swagger.io/v2/schema.json)
 
-Files referenced to an URL will automatically be cached if the first path in
-L</cache_paths> is writable.
+Files referenced to an URL will automatically be cached if L</cache_paths> is supplied.
+Note that the Cache-Control header for remote files is B<not> honored so you will
+manually need to remove any cached remote files should you need to refresh them.
```

t/gh-42-cache-control.t
```diff
@@ -0,0 +1,36 @@
...
+ok(@old_files == @new_files, 'remote file not cached in default cache dir');
+
+my $tmp_dir = File::Temp->newdir;
+$validator = JSON::Validator->new(cache_paths => [@{$validator->cache_paths}, $tmp_dir->dirname]);
+
+$validator->schema('https://za.payprop.com/api/docs/api_spec.yaml');
+isa_ok($validator->schema, 'Mojo::JSON::Pointer');
+
+@new_files = _get_cached_file_paths($validator);
+ok(@new_files > @old_files, 'remote file cached when cache_paths not the default');
```
jhthorsen: I think this test is way better than mine. (#45)

leejo:
> Please have a look at #45 as well. I think I will merge your PR and apply some changes on top. I will leave all of these open until we have discussed it further.

Sounds good.

jhthorsen:
The merge did not turn out quite as expected: I moved some documentation around. Please let me know if the result is not good enough.

PR #49, opened 2017-02-24
Allow alternative date-time separator

ugexe:
RFC3339 optionally allows the date and time separator 'T' to be
anything. This changes the previous date-time regex to the one
used by Mojo::Date, so cases like '2014-12-09 20:49:37Z' are
not rejected.

See: https://github.com/kraih/mojo/blob/master/lib/Mojo/Date.pm#L9
See: https://github.com/jhthorsen/json-validator/issues/48

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -719,8 +719,12 @@ sub _guessed_right {
 }
 
 sub _is_date_time {
-  $_[0] =~ qr/^(\d+)-(\d+)-(\d+)T(\d+):(\d+):(\d+(?:\.\d+)?)(?:Z|([+-])(\d+):(\d+))?$/io;
+  $_[0] =~ qr/
+    ^(\d+)-(\d+)-(\d+)\D+(\d+):(\d+):(\d+(?:\.\d+)?)   # Date and time
+    (?:Z|([+-])(\d+):(\d+))?$                          # Offset
+  /xi;
 }
+
 sub _is_domain { warn "Data::Validate::Domain is not installed"; return; }
 
 sub _is_email {
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -720,8 +720,8 @@ sub _guessed_right {
 
 sub _is_date_time {
   $_[0] =~ qr/
-    ^\d+-\d+-\d+\D+\d+:\d+:\d+(?:\.\d+)?   # Date and time
-    (?:Z|[+-]\d+:\d+)?$                    # Offset
+    ^\d{4}-\d{2}-\d{2}.\d{2}:\d{2}:\d{2}(?:\.\d+)? # Date and time
+    (?:Z|[+-]\d{2}:\d{2})?$                        # Offset
   /xi;
 }
 
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -720,8 +720,8 @@ sub _guessed_right {
 
 sub _is_date_time {
   $_[0] =~ qr/
-    ^(\d+)-(\d+)-(\d+)\D+(\d+):(\d+):(\d+(?:\.\d+)?)   # Date and time
-    (?:Z|([+-])(\d+):(\d+))?$                          # Offset
+    ^\d+-\d+-\d+\D+\d+:\d+:\d+(?:\.\d+)?   # Date and time
+    (?:Z|[+-]\d+:\d+)?$                    # Offset
   /xi;
 }
 
--- a/t/jv-formats.t
+++ b/t/jv-formats.t
@@ -10,6 +10,10 @@ my @errors;
   $schema->{properties}{v}{format} = 'date-time';
   @errors = $validator->validate({v => '2014-12-09T20:49:37Z'}, $schema);
   is "@errors", "", "date-time valid";
+  @errors = $validator->validate({v => '2014-12-09 20:49:37Z'}, $schema);
+  is "@errors", "", "date-time with alternate separator valid";
+  @errors = $validator->validate({v => '2014-12-09T20:49:37.00Z'}, $schema);
+  is "@errors", "", "date-time with fractional seconds valid";
   @errors = $validator->validate({v => '20:46:02'}, $schema);
   is "@errors", "/v: Does not match date-time format.", "date-time invalid";
 }


tyldum:
Just for the record:
RFC3339 explicitly forbids anything but "T". ISO8601 allows "T" to be omitted for the sake of readability, which I would argue is not the case for JSON.
I got no strong opinions on it, but just wanted the records to be correct.

ugexe:
[RFC3339](https://tools.ietf.org/html/rfc3339#section-5.6)
```
      NOTE: Per [ABNF] and ISO8601, the "T" and "Z" characters in this
      syntax may alternatively be lower case "t" or "z" respectively.

      This date/time format may be used in some environments or contexts
      that distinguish between the upper- and lower-case letters 'A'-'Z'
      and 'a'-'z' (e.g. XML).  Specifications that use this format in
      such environments MAY further limit the date/time syntax so that
      the letters 'T' and 'Z' used in the date/time syntax must always
      be upper case.  Applications that generate this format SHOULD use
      upper case letters.

      NOTE: ISO 8601 defines date and time separated by "T".
      Applications using this syntax may choose, for the sake of
      readability, to specify a full-date and full-time separated by
      (say) a space character.
```

tyldum:
I think appendix A further specifies how to interpret the ISO quotation.
But it's not as crystal clear as it should be.

ugexe:
[Appendix A. ISO 8601 Collected ABNF](https://tools.ietf.org/html/rfc3339#page-12)
```
Note that due to ambiguities in ISO 8601, some interpretations had to be made.

...

ISO 8601 states that the "T" may be omitted under some circumstances.  This grammar requires the "T" to avoid ambiguity.
```
The appendix states towards the top that the grammar used to defined ISO8601 is not complete, and lists some of its caveats (including the date and time separator)

tyldum:
Googling shows this as a common topic in many date-time libraries, with various outcomes. The majority forbids space.
Tricky, especially for a validator :)

lib/JSON/Validator.pm
```diff
@@ -719,8 +719,12 @@ sub _guessed_right {
 }
 
 sub _is_date_time {
-  $_[0] =~ qr/^(\d+)-(\d+)-(\d+)T(\d+):(\d+):(\d+(?:\.\d+)?)(?:Z|([+-])(\d+):(\d+))?$/io;
+  $_[0] =~ qr/
+    ^(\d+)-(\d+)-(\d+)\D+(\d+):(\d+):(\d+(?:\.\d+)?)   # Date and time
```
jhthorsen: Two things:

* While you're making this change, could you remove all the captures? There's no need for it here, since it's just a validator.
* I'm pretty sure `\D+` is incorrect. If it indeed allows any character to be used as separator it should just be `.`.

jhthorsen:
Could you let me know what impact (good or bad) accepting this PR will have for you?

From how I read the spec, it's rather silly to write "data-time" with a space, but it does not seem to be illegal.

lib/JSON/Validator.pm
ugexe: I implemented a more conservative option, although I also considered `.`. I didn't choose `.` because:
1) I'm unsure if the wording means that the separator can be omitted entirely (and not just replaced)
2) If allowing a digit as a separator there is in the spirit of the RFC itself

ugexe:
For the record:
````
$ date --rfc-3339=ns
2017-02-25 12:09:40.629915598-05:00
```

lib/JSON/Validator.pm
jhthorsen: Omitted? "." means that a character is required, but it can be anything. `\D+` does also not mean it can be omitted. It will match "xxxxx", "x", "T" or a space - not conservative at all to me. I will not take this PR with `\D+`. I think we need to choose between:
* `.`: Exactly one character, which can be anything. (This is how I read the spec)
* `[Tt\s]`: Something that is more sane to me.

jhthorsen:
I'm still not sure about the regex. I think we should simply vote. Appreciate votes from anyone who wants.

leejo:
> I'm still not sure about the regex.

I'm fairly agnostic but given how hand wavy the RFC is about this it's probably just going to come up again and again so think it makes sense purely from that point of view. The alternative is people just end up forking it and we end up with JSON::Validator::Lax.

The change here doesn't update the perldoc so it needs that as well.

tyldum:
If any character is allowed, then you will have a ::Strict fork instead. I see the point of being liberal in what you receive, but is that the role of a validator? I really don't know. I am fine with a single space as option, but definitely not any character. I use a validator to restrict the amount of trickery needed later on, especially for dates. On Feb 25, 2017 22:24, Lee J <notifications@github.com> wrote:@leejo commented on this pull request.

Perldoc needs updating as it currently states:
An RFC3339 timestamp in UTC time. This is formatted as "YYYY-MM-DDThh:mm:ss.fffZ". The milliseconds portion (".fff") is optional

This is now not quite true with this change. And the regexp isn't quite strict enough if we're allowing just a space or the T char?

ugexe:
Is this too strict? It only allows T and space (and t via `/i`) as requested, but additionally validates most of the range restrictions:
```
  $_[0] =~ qr/
    ^\d{4}-(?:0[1-9]|1[012])-(?:0[1-9]|[12][0-9]|3[01])           # Date
    [T\s]                                                         # Date and time separator
    (?:[01][0-9]|2[0-3]):(?:[0-5][0-9]):(?:[0-5][0-9]|60)(\.\d+)? # Time with any number of optional fractional seconds
    (?:Z|[+-](?:[01][0-9]|2[0-3]):[0-5][0-9])$                    # Offset
  /xi;
```

jhthorsen:
@ugexe: The best way to check if it's too strict or not is to steal the test suite from another library and check if JSON::Validator accepts the same input.

ugexe:
@jhthorsen Sorry I meant is it ok to add this additional (valid) stricter logic, and not if the logic I wrote is technically too strict.

jhthorsen:
It's ok as long as it's more correct.

mrenvoize:
I think ugexe's regex is technically the most correct to the RFC. It would get my vote :)

jhthorsen:
@ugexe: I like your regex, but it requires a bunch of unit tests. Please update the PR and _at least_ test with these dates:

1999-12-31T00:00:00Z
2222-01-01T13:50:05+01:00
2002-01-01 23:59:59+12:00
2099-07-29T13:50:05-05:00
1880-01-01 04:56:02Z
9999-13-11 04:56:02Z
9999-12-00 04:56:02-00:00
9999-12-00 25:99:02Z
9999-12-01 12:12:60Z

(Note that some of them are failing)

I'm also not sure if this is the way to solve the problem if we want to be more strict. How about capturing the data and check if it's in between some given limits? Something like this:

```
sub _is_date_time {
  return undef unless my @date = $_[0] =~ qr/
    ^
    (\d{4})-(\d{2})-(\d{2})           # Date
    [T\s]
    (\d{2}):(\d{2}):(\d{2})(?:\.\d+)? # time
    (?:Z|[+-]\d{2}:\d{2})?            # Offset
    $
  /xi;

  return undef if $date[1] > 12;
  return undef if $date[2] > 31;
  return undef if $date[3] > 23;
  return undef if $date[4] > 59;
  return undef if $date[5] > 59;
  return 1;
}
```

mrenvoize:
That's a solid set of test dates you've added there @jhthorsen!

I hadn't thought of changing from a regex to a regex + capture tests.. would be really interested to see a little benchmark comparing pure regex (as cryptic to read as it may be) vs regex with captures and gt tests.

I'll go create a benchmark to satisfy my own curiosity

tyldum:
Any checks beyond the basics (like ranges) should be left to a library. It will break. Time is one of those things that appear to be simple, but are not.

jhthorsen:
@tyldum: I agree. Here is my suggestion:

```
use Time::Local ();

sub _is_date_time {
  my @time = $_[0]
    =~ m!^(\d{4})-(\d\d)-(\d\d)[T\s](\d\d):(\d\d):(\d\d(?:\.\d+)?)(?:Z|([+-])(\d+):(\d+))?$!io;
  return 0 unless @time;
  local $@;
  @time = map { s/^0//; $_ } reverse @time[0 .. 5];
  $time[4] -= 1;
  return eval { Time::Local::timegm(@time); 1 } || 0;
}

# tests
use Test::More;
sub check_date_time { is(_is_date_time($_[0]), $_[1], $_[0]); }

check_date_time("2017-03-29T23:02:55.831Z",  1);
check_date_time("2017-03-29t23:02:55.01z",   1);
check_date_time("2017-03-29 23:02:55-12:00", 1);
check_date_time("2016-02-29T23:02:55+05:00", 1);

check_date_time("xxxx-xx-xxtxx:xx:xxz", 0);
check_date_time("2017-03-29T23:02:60Z", 0);
check_date_time("2017-03-29T23:61:55Z", 0);
check_date_time("2017-03-29T24:02:55Z", 0);
check_date_time("2017-03-32T23:02:55Z", 0);
check_date_time("2017-02-30T23:02:55Z", 0);
check_date_time("2017-02-29T23:02:55Z", 0);
check_date_time("2017-13-29T23:02:55Z", 0);
check_date_time("2017-03-00T23:02:55Z", 0);
check_date_time("2017-00-29T23:02:55Z", 0);

done_testing;
```

jhthorsen:
Any comments on my suggestion?

jhthorsen:
I will revert 984aec0 unless I get feedback on it. Please try it out!

    cpanm https://github.com/jhthorsen/json-validator/archive/master.tar.gz

mrenvoize:
Works well for me, currently running with this via OpenAPI in my app... plenty of date-time fields coming in and out including some one's which are now failing correctly on bad inbound requests :)

PR #52, opened 2017-02-28
_resolve_schema sometimes unintentionally skips an unresolved $ref

taskula:
After migrating from Swagger2 to Mojolicious::Plugin::OpenAPI, we noticed that every now and then the requests to our API would fail in what seems to be a pretty random order. We figured out it has to do with $refs sometimes being unresolved.

The issue seems to occur in _resolve_schema which unintentionally skips an unresolved $ref. In order to avoid resolving same references twice or more, _resolve_schema has a functionality that stores the memory address of the reference to remember that it has already resolved it, but it seems that at some point that exact same memory address is freed for another reference and it would then get skipped.

I have created a test that replicates the original issue with our spec. I included our whole spec because I was not able to reproduce it with a smaller specification... Perhaps someone can come up with a better test and the first commit containing our whole API could be ignored? Please see the commit messages for more documentation. Thanks!

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -258,7 +258,10 @@ sub _resolve_schema {
 
         # int($v) returns the memory address of the reference. The "if" below
         # will not resolve the same "$ref" over again.
-        next if ref $v and $self->{seen}{int($v)}++;
+        next if ref $v and $self->{seen}{int($v)};
+        # Store the reference to keep that memory address in use for that
+        # particular reference.
+        $self->{seen}{int($v)} = $v if ref $v;
 
         # Make sure we do not modify the input data structure.
         # Changing the input makes t/expand.t in swagger2.git fail.
--- a/t/openapi-unresolved-ref.t
+++ b/t/openapi-unresolved-ref.t
@@ -1,0 +1,34 @@
+use Mojo::Base -strict;
+use Test::More;
+use JSON::Validator::OpenAPI;
+use Data::Dumper;
+
+#for my $i (1..20) {
+my $validator = JSON::Validator::OpenAPI->new->load_and_validate_spec(
+    't/spec/koha-swagger/swagger.json',
+    {
+        allow_invalid_ref => 1
+    }
+);
+
+is(for_hash($validator->schema->{data}), 0, 'No $refs unresolved.');
+#}
+
+sub for_hash {
+    my ($hash) = @_;
+    while (my ($key, $value) = each %$hash) {
+        if ('HASH' eq ref $value) {
+            my $ret = for_hash($value);
+            if ($ret) {
+                return $ret;
+            }
+        }
+        else {
+            if ($key eq '$ref') {
+                return $value;
+            }
+        }
+    }
+}
+
+done_testing;
--- a/t/spec/koha-swagger/definitions.json
+++ b/t/spec/koha-swagger/definitions.json
@@ -1,0 +1,59 @@
+{
+  "accountline": {
+    "$ref": "definitions/accountline.json"
+  },
+  "availability": {
+    "$ref": "definitions/availability.json"
+  },
+  "availability/biblio": {
+    "$ref": "definitions/availability/biblio.json"
+  },
+  "availability/item": {
+    "$ref": "definitions/availability/item.json"
+  },
+  "city": {
+    "$ref": "definitions/city.json"
+  },
+  "patron": {
+    "$ref": "definitions/patron.json"
+  },
+  "patronstatus": {
+    "$ref": "definitions/patronstatus.json"
+  },
+  "session": {
+    "$ref": "definitions/session.json"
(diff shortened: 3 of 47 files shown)


jhthorsen:
The test case is huge. I'm pretty sure we can replicate this with a smaller test case while setting PERL_HASH_SEED. https://github.com/jhthorsen/json-validator/blob/master/test.sh#L25

I'm assigning this to @mrenvoize, because he had another bug which this fix will re-introduce. The change that makes this fail is f17287ee0063ddc5d9594804fecde2d05f24fc13.

taskula:
Yes, that particular test case should not be merged. But I'm not really familiar with PERL_HASH_SEED either. A helping hand with creating a proper test is much appreciated.

I'm not sure how this change re-introduces the recursion bug? At least the recursion test was passing after the change. I intended no change in functionality, but instead only making sure that the memory address stays for that particular reference as it now looks like it is somehow freed.

mrenvoize:
Indeed, I'm not sure it re-introduces the recursion issue in the form taskula has resolved the issue.. However, I'm not entirely sure that form is the best form.. mostly due to comments on IRC

lib/JSON/Validator.pm
```diff
@@ -258,7 +258,10 @@ sub _resolve_schema {
 
         # int($v) returns the memory address of the reference. The "if" below
         # will not resolve the same "$ref" over again.
-        next if ref $v and $self->{seen}{int($v)}++;
+        next if ref $v and $self->{seen}{int($v)};
```
jhthorsen: Not sure if this is a crazy idea or not, but how about doing this:

```
next if ref $v and $self->{seen}{_md5($v)};
...
sub _md5 { Mojo::Util::md5_sum(Mojo::JSON::encode_json($v)) }
```

mrenvoize: That's a really interesting alternative as it removes the reliance on memory addresses entirely.

I'm still interested to see how we manage to get into this position in the first place.. that still perplexes me :(

taskula: This solution seems to fix my original issue and at the moment I don't have anything against this change.

kivilahtio:
Hi!

Can we get this fix pushed? We have spent quite a few days working on this and the test case is impossible to set up.

We are bleeding here, really.

jhthorsen:
@kivilahtio: There's no suggested fix that I'm too keen on implementing. I'm not sure, but maybe this will fix the issue, without too many ugly hacks:

https://github.com/jhthorsen/json-validator/compare/bug/acceptance-t?expand=1#diff-adc6881507836d26fecdde466d40eea1R276

You can try it locally with this command (even if it's not done):

    $ cpanm -n https://github.com/jhthorsen/json-validator/archive/bug/acceptance-t.tar.gz

I need too remind everybody, that I need a minimal test cast to fix this.

mrenvoize:
Hmm, Not sure I fully understand the changes in that last commit you've referenced there @jhthorsen.

I'll have a play with master over the next couple of days.. though I wasn't able to replicate the problem in the first place outside of the massive test schema attached to this bug :(.

I would still like to see a more minimal test case for this bug.

jhthorsen:
There are some changes that are important:

* New logic for tracking "seen", copied from Data::Dumper: https://github.com/jhthorsen/json-validator/commit/dc9b327e559c38806b3e122f4bdea6e098845181#diff-adc6881507836d26fecdde466d40eea1R313
* Tracking "seen" on a higher level. Not sure if it matters, but it might avoid "loosing track" for the ref: https://github.com/jhthorsen/json-validator/commit/dc9b327e559c38806b3e122f4bdea6e098845181#diff-adc6881507836d26fecdde466d40eea1R266
* Using "unshift" will result in "deepest first" traversing: https://github.com/jhthorsen/json-validator/commit/0759e228f2bc746947cf390c2967106f961ec3fb#diff-adc6881507836d26fecdde466d40eea1R270
* I don't care if I modify the input structure anymore: https://github.com/jhthorsen/json-validator/commit/0759e228f2bc746947cf390c2967106f961ec3fb#diff-adc6881507836d26fecdde466d40eea1L279
* Continue tracking refs when copied in: https://github.com/jhthorsen/json-validator/commit/0759e228f2bc746947cf390c2967106f961ec3fb#diff-adc6881507836d26fecdde466d40eea1R306

The reason why I think these changes also affect this PR, is that there was some failing tests in acceptance.t that seems related, which are now fixed.

```
git checkout 7979140
prove -l t/acceptance.t
```

taskula:
It appears 0759e22 affects this PR and fixes my original issue. Sorry for failing to provide a minimal test case. I'm ok with closing this.

mrenvoize:
All tests are passing for me locally.. and there's no issues when using in my app either.. gets the go ahead from me :)

jhthorsen:
Awesome! Will make a new release later today.

Issue #53, opened 2017-02-28
Same file loaded multiple times


taskula:
Commit 12808e9a9ac4978f065cfac4ae35afad9b7f77dd replaces Cwd::abs_path() with Mojo::File->to_abs(). After https://github.com/kraih/mojo/issues/1030, Mojo::File->to_abs has been using File::Spec's rel2abs. rel2abs documentation mentions:

> If $path is absolute, it is cleaned up and returned using "canonpath".

Then, canonpath has the following documentation

> Note that this does *not* collapse x/../y sections into y. This is by design. If /foo on your system is a symlink to /bar/baz, then /foo/../quux is actually /bar/quux, not /quux as a naive ../-removal would give you. If you want to do this kind of processing, you probably want Cwd's realpath() function to actually traverse the filesystem cleaning up paths like this.

Cwd::realpath() is a synonym for Cwd::abs_path(). This means that since 12808e9a9ac4978f065cfac4ae35afad9b7f77dd, we are no longer cleaning up the paths the same way as with Cwd::abs_path(). In certain circumstances, this will cause the same file to be loaded multiple times because the path is represented in different ways.

For example in my spec I now see the same file being loaded twice:
```
[JSON::Validator] Loading schema /home/ubuntu/json-validator/t/spec/koha-swagger/definitions.json namespace=/home/ubuntu/json-validator/t/spec/koha-swagger/definitions.json scheme=file
[JSON::Validator] Loading schema /home/ubuntu/json-validator/t/spec/koha-swagger/paths/../definitions.json namespace=/home/ubuntu/json-validator/t/spec/koha-swagger/paths/../definitions.json scheme=file
```
(...and the same is happening for multiple different files)

When I replace Mojo::File->to_abs back to Cwd::abs_path at https://github.com/jhthorsen/json-validator/commit/12808e9a9ac4978f065cfac4ae35afad9b7f77dd#diff-adc6881507836d26fecdde466d40eea1L132, the same files are no longer loaded multiple times and it shows as a nice performance improvement (about half a second for our relatively large spec).

Would it be reasonable to switch back to Cwd::abs_path() or perhaps come up with another solution to avoid loading the same file multiple times? 

Thanks!

jhthorsen:
Yes, absolutely. Thanks for the bug report. Do you have time to write a failing test and/or submit a PR, where abs_path() is used? I need a failing test, so I don't introduce the same issue again.

taskula:
[...]

PR #54, opened 2017-03-01
Avoid loading the same file multiple times

taskula:
See issue #53.


mrenvoize:
Looks reasonably to me. How keen are you on keeping to Mojo::File batman.. is it perhaps something Mojo::File aught to 'get right'?

mrenvoize:
Looks like that may have already been answered in the original issue, switching back to CWD seems ok for this.

PR #57, opened 2017-03-17
Replace _uniq function with List::MoreUtils#uniq

akzhan:
(it's embedded with Perl and provide XS implementation whenever possible).

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -15,6 +15,6 @@ WriteMakefile(
     },
   },
   BUILD_REQUIRES => {'Test::More'  => '0.88', 'Test::Warnings' => '0.016'},
-  PREREQ_PM      => {'Mojolicious' => '7.15'},
+  PREREQ_PM      => {'Mojolicious' => '7.15', 'List::MoreUtils' => undef},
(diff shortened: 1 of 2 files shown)


jhthorsen:
List::MoreUtils is not a core module I'm afraid.

Thanks for showing interest!

akzhan:
Yes, I made mistake.

You may use `use List::Util 1.45 qw( uniq );` but bundled version may be less than required 1.45.

jhthorsen:
What benefit does this change have?

jberger:
Note that you also have http://mojolicious.org/perldoc/Mojo/Collection#uniq at your disposal.

akzhan:
Nothing but less hand-made bicycles.

speed and memory consumption doesn't matter in your case.

jberger:
I would recommend staying far away from List::MoreUtils at present. If speed/memory are required I'd use List::Util. That said, I doubt that that is high in the memory footprint nor the speed profile.

jhthorsen:
I'm locking this issue now. You're welcome to throw bad comments on us in #swagger on irc.perl.org @akzhan.

PR #60, opened 2017-04-06
Replace Windows path separators with Unix path separators

charsbar:
... so that the separators will not be URL-escaped afterwards, which then will let Cwd::abs_path crash (line 144), saying there's no such file or directory.

See http://www.cpantesters.org/cpan/report/ed0d39b3-6bfc-1014-941f-c7baf1edae7c for crash example.


jhthorsen:
Thanks! 👍

PR #62, opened 2017-05-19
implement "const"

kgoess:
added in draft-wright-json-schema-validation-01

http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.24

I wasn't sure what the intent of the error-handling logic was in _validate_type_object, so I just copied the existing code covering "enum". Feel free to correct that or clean it up.


jhthorsen:
The code is not tidied, but looks good otherwise. I will merge it and do some clean up.

Thanks! 👍

PR #64, opened 2017-06-05
Don't checks data type, if data undefined

Pilat66:
I suggest not checking the data type if the data contains undefined

For schema
```
Family:
    type: object
    properties:
        id: 
            type: integer
        name: 
            type: string
        subname: 
            type: string
      required:
      - id
      - name
      - subname
```
does not pass validation:
```
{
    id => 1,
    name=>'name',
    subname=>undef,
}
```

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -340,10 +340,15 @@ sub _validate {
     push @errors, $self->_validate_any_of($data, $path, [map { +{%$schema, type => $_} } @$type]);
   }
   elsif ($type) {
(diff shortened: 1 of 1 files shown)


jhthorsen:
I will not take this PR, since what you're suggesting is not correct.

If you want `undef` to be accepted, then you need to specify that in the schema, like this:

```
"subname": {
  "type":["string","null"]
}
```

Pilat66:
"null" not supported on swaggerhub; last update Eclipse plugin also mark it as error.
As a result, the version of OpenAPI 3 is not supported, and version 2 is supported in a state where it is very difficult to edit.

Issue #56, opened 2017-03-06
Make boolean coercion work for B::SVp_IOK values


fabzzap:
There's a line in `_validate_type_boolean`

    and (B::svref_2object(\$value)->FLAGS & B::SVp_NOK or $value =~ /^(true|false)$/))

so that values that have the flag B::SVp_NOK on are coerced but other values are not. In other parts of the code, `B::svref_2object(\$value)->FLAGS & (B::SVp_IOK | B::SVp_NOK)` is checked instead, and this makes more sense. Now, with coerce on, 0.5 is considered a valid boolean value but 2 is not, for example.

Why not change `_validate_type_boolean` so  it checks `B::SVp_IOK | B::SVp_NOK`?

jhthorsen:
Good question. Not sure why it doesn't coerce all values, but the test I cleaned up in 0abd552 shows how I thought it should work, but I can't remember why right now 😞

jhthorsen:
Could you write some tests and make a PR? I have a hard time seing how this will actually look.

jhthorsen:
Closing this because of lack of interest and I don't have time to look at it myself.

PR #67, opened 2017-06-18
Coerce integer numbers into booleans

fabzzap:
This intends to solve issue #56 , so that, when coerce is on, integer numbers are coerced into booleans, just like real numbers are currently.

It also adds a test case for boolean types. Currently master fails one test case (when coerce is on and the value is 1000), the code in this pull request succeeds

### DIFF
--- a/t/jv-boolean.t
+++ b/t/jv-boolean.t
@@ -1,15 +1,12 @@
 use lib '.';
 use t::Helper;
-use Test::More;
-use utf8;
-use JSON;
 
 my $schema = {
   type       => 'object',
   properties => {nick => {type => 'boolean'}}
 };
 
-validate_ok {nick => JSON::true}, $schema;
+validate_ok {nick => true}, $schema;
 validate_ok {nick => 1000},       $schema, E('/nick', 'Expected boolean - got number.');
 validate_ok {nick => 0.5},        $schema, E('/nick', 'Expected boolean - got number.');
 validate_ok {nick => 'nick'},     $schema, E('/nick', 'Expected boolean - got string.');
--- a/t/jv-boolean.t
+++ b/t/jv-boolean.t
@@ -1,0 +1,31 @@
+use lib '.';
+use t::Helper;
+use Test::More;
+use utf8;
+use JSON;
+
+my $schema = {
+  type       => 'object',
+  properties => {nick => {type => 'boolean'}}
+};
+
+validate_ok {nick => JSON::true}, $schema;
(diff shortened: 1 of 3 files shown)


t/jv-boolean.t
```diff
@@ -0,0 +1,31 @@
+use lib '.';
+use t::Helper;
+use Test::More;
+use utf8;
+use JSON;
```
jhthorsen: Please don't depend on "JSON.pm" (since it's not a core module). Do this instead:

```
use lib;
use t::Helper; # <-- this imports Test::More and utf8
```

t/jv-boolean.t
```diff
@@ -0,0 +1,31 @@
...
+
+my $schema = {
+  type       => 'object',
+  properties => {nick => {type => 'boolean'}}
+};
+
+validate_ok {nick => JSON::true}, $schema;
```
jhthorsen: t::Helper exports `true()`, so you can do this:

```
validate_ok {nick => true}, $schema;
```

t/jv-boolean.t
```diff
@@ -0,0 +1,31 @@
...
+validate_ok {nick => 1000},       $schema, E('/nick', 'Expected boolean - got number.');
+validate_ok {nick => 0.5},        $schema, E('/nick', 'Expected boolean - got number.');
+validate_ok {nick => 'nick'},     $schema, E('/nick', 'Expected boolean - got string.');
+validate_ok {nick => bless({}, 'BoolTest')},   $schema;
+validate_ok {nick => bless({}, 'BoolTestNegative')},   $schema, E('/nick', 'Expected boolean - got BoolTestNegative.');
```
jhthorsen: What's the reason for testing both "BoolTest" and "BoolTestNegative". What's the difference between the two tests?

jhthorsen:
Note to myself: After merging this, move the boolean tests from `t/jv-basic.t` into this test, and (maybe) rename `t/jv-basic.t`to `t/jv-null.t`.

t/jv-boolean.t
fabzzap: I pushed a commit to my fork to fix this and the above

t/jv-boolean.t
fabzzap: BoolTest stringifies to 1 and therefore validates OK. BoolTest stringifies to 2 and therefore validates with an (expected) error. Blessed references validate OK if the stringify to 1 or anything false

    return if Scalar::Util::blessed($value) and ("$value" eq "1" or !$value);

jhthorsen:
LGTM! Thank you :

Issue #65, opened 2017-06-12
json reference validation


ckongEbi:
Hi, can I know if `json reference` is validated ? Since by running the validator, we realised that it does not complain, even though the reference is non-existent. 

https://github.com/opentargets/json_schema/blob/master/src/evidence/literature_mining.json#L27

`https://raw.githubusercontent.com/opentargets/json_schema/1.2.5/src/evidence/base.json#base_evidence/definitions/single_lit_reference`

"base_evidence" doesn't exisit
https://github.com/opentargets/json_schema/blob/master/src/evidence/base.json#L80
`"$ref": "#/definitions/single_lit_reference"`

Thanks!
ck

jhthorsen:
I don't understand how that's possible. Please provide a small test. This example is way to big for me to wrap my head around. Including refs from external documents are tested, but maybe there's a missing failing test.

ckongEbi:
Hi, I ran with the script, which didn't complain about the missing reference, might I be missing something? thanks!

`perl json_schema_validator.pl test.txt \
 https://raw.githubusercontent.com/opentargets/json_schema/master/src/literature_mining.json
3 evidence PASSED schema validation!
`
Script here:
`https://github.com/opentargets/json_schema/blob/master/scripts/json_schema_validator.pl`

Example json:
[test.txt](https://github.com/jhthorsen/json-validator/files/1068352/test.txt)

jhthorsen:
That's not a small test case. This is a small test case: https://github.com/jhthorsen/json-validator/blob/master/t/spec/with-relative-ref.json

jhthorsen:
Also, what is `json_schema_validator.pl` ?

jhthorsen:
Just ran your script and got the errors below, but I have no idea if that's related to your issue or not. Again: Please provide a smaller test case. You can see examples here: 

* https://github.com/jhthorsen/json-validator/blob/master/t/relative-ref.t
* https://github.com/jhthorsen/json-validator/blob/master/t/invalid-ref.t

```
perl json_schema_validator.pl ~/Downloads/test.txt https://raw.githubusercontent.com/opentargets/json_schema/master/src/literature_mining.json
Validation error on evidence line 1 : 	$VAR1 = bless( {
                 'path' => '/',
                 'message' => 'allOf failed: allOf failed: oneOf failed: Properties not allowed: value. Properties not allowed: value. Properties not allowed: value. Properties not allowed: value.'
[...]

PR #75, opened 2017-10-04
When caching / resolving a schema avoid the case where $namespace is just a fragment  [ Fixes issue #65 ]

wbazant:
This reinstates the support of remote schemas where every document has an id that is just a fragment.

I ran the test suite on a variant of this with debugger statements, https://github.com/wbazant/json-validator/commit/184bcda2810fbace8f05ae9e35f71220b18f1a5e , and logs said the change didn't touch execution paths covered by the suite apart from the two new tests.
Without the head commit, the first test succeeds and the second one fails. In 1.0.3, they both fail.

It won't be helpful until resolving https://github.com/jhthorsen/json-validator/issues/74 and bringing back support for remote schemas that have ids - and if it adds support for remote schemas that have ids that also reference each other, this PR would have become obsolete.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -323,7 +323,9 @@ sub _seen {
 
 sub _store {
   my ($self, $key, $namespace, $schema) = @_;
-  $namespace = Mojo::URL->new($namespace)->fragment(undef)->to_string;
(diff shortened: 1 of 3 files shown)


jhthorsen:
I'm closing this in favour of #79.

PR #76, opened 2017-10-06
Better booleans handling in OpenAPI

khvzak:
If a user passes invalid boolean values in request, no errors will be generated.

For instance, for the schema:
```
parameters => [
  {name => "param", type => 'boolean', in => 'query'},
],
```
And requests:
`GET /index?param=FALSE`
or
`GET /index?param=INVALID`

the `param` will be provided to a controller with "true" value, which is not correct.

With this PR it will be correct only with boolean's coercion turned on.

### DIFF
--- a/lib/JSON/Validator/OpenAPI.pm
+++ b/lib/JSON/Validator/OpenAPI.pm
@@ -83,7 +83,12 @@ sub validate_request {
         $value += 0;
       }
       elsif ($type eq 'boolean') {
-        $value = (!$value or $value eq 'false') ? Mojo::JSON->false : Mojo::JSON->true;
+        if (!$value or $value =~ /^(?:false|off)$/) {
+          $value = Mojo::JSON->false;
+        }
+        elsif ($value =~ /^(?:1|true|on)$/) {
+          $value = Mojo::JSON->true;
+        }
       }
     }
 
--- a/t/openapi-request.t
+++ b/t/openapi-request.t
@@ -138,6 +138,34 @@ validate_request(
   }
 );
 
+# valid booleans
+$c->tx->req(Mojo::Message::Request->new)->req->url->parse('http://127.0.0.1/?a=&b=0&c=false&x=true&y=1&z=on');
+validate_request(
+  {
+    parameters => [
+      map { +{name => $_, type => 'boolean', in => 'query'} } qw/a b c x y z/,
+    ],
+  },
+  sub {
+    is "@_", "", "valid request";
+    is_deeply $input, {a => false, b => false, c => false, x => true, y => true, z => true}, 'valid booleans';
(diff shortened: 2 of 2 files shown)


lib/JSON/Validator/OpenAPI.pm
```diff
@@ -83,7 +83,12 @@ sub validate_request {
         $value += 0;
       }
       elsif ($type eq 'boolean') {
-        $value = (!$value or $value eq 'false') ? Mojo::JSON->false : Mojo::JSON->true;
+        if (!$value or $value =~ /^(?:false|off)$/) {
+          $value = Mojo::JSON->false;
+        }
+        elsif ($value =~ /^(?:1|true|on)$/) {
```
jhthorsen: Where does “on” and “off" come from?

Pilat66: Probably, this is one of the versions of true / false representation in some system. If so, it would be nice to add 't' / 'f' - in PostgreSQL it means true / false and I have to manually translate these values. But probably this should be moved to _validate_type_boolean () and change the conversion to boolean.

jhthorsen:
I dropped support for "off" and "on". Please refer to other libraries or some sort of common specification to get support for other boolean strings.

khvzak:
@jhthorsen 
"on" is a default value for html checkbox element.
For instance:
```
<html>
<body>
  <form method=GET>
    <input name=boolean type=checkbox />
    <input type=submit />
  </form>
</body>
</html>
```
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value

"off" is just opposite and isn't specificated.
But it's up to you, is it make sense to add support for "on" or not.

jhthorsen:
I don't really want to support on/off. But if enough people disagree, then I might change my mind.

perlpunk:
From my point of view, if you want to take the contents of an HTML form to submit it to a REST API, then you should make sure you submit the supported values.
In the given example, that's really easy by adding `value="true"` to the checkbox.
Also, from the example I don't see why "off" should be supported, just because it's the opposite.

Pilat66:
JSON, passed validation from 1/0, on/off, t/f, will not later be parsed as a JavaScript structure, that is, it will not be JSON.
But the conversion of 1/0, on/off, t/f to true / false, which you can do knowing that there must be a Boolean value in the source data, must be performed with the modification of the source data.
In other words, _validate_type_boolean () should not only return true, but also modify the value in the structure that is checked.

jhthorsen:
@perlpunk: I agree.

@Pilat66: I don't follow. This "on/off" thing would only be done for things like query strings or headers, so adding this to _validate_type_boolean() is not going to happen.

PR #77, opened 2017-10-06
Validate headers regardless of case

khvzak:
HTTP headers are defined as case-insensitive and probably should be validated using this rule.

RFC 7230 (https://tools.ietf.org/html/rfc7230#section-3.2) 
RFC 2616 (https://tools.ietf.org/html/rfc2616#section-4.2)

### DIFF
--- a/lib/JSON/Validator/OpenAPI.pm
+++ b/lib/JSON/Validator/OpenAPI.pm
@@ -53,6 +53,8 @@ sub validate_request {
   my ($self, $c, $schema, $input) = @_;
   my @errors;
 
+  my $headers; # for caching
+
   for my $p (@{$schema->{parameters} || []}) {
     my ($in, $name, $type) = @$p{qw(in name type)};
     my ($exists, $value);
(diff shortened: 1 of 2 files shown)


jhthorsen:
Awesome! I made the change myself, because of the conflict I created when closing #76.

khvzak:
@jhthorsen 

Thanks for submitting the changes.
I just worried that the total complexity of the 2nd (submitted) variant to validate headers is `O(M*N)` where M - number of defined headers in a schema, and N is total number of headers.

The 1st variant has complexity `O(M + N)`.

jhthorsen:
I don't follow. Can you comment on the commit?

khvzak:
For every parameter (with `in:header`) defined in schema, we executing the following code:
```
sub validate_request {
   ...
  for my $p (@{$schema->{parameters} || []}) {
    ...
    $value  = $self->_get_request_data($c, $in);
    ...
  }
   ...
}
```
And the function `_get_request_data` every time recalculate lower-case header's hash:
```
my $headers = $c->req->headers->to_hash(1);
return {map { lc($_) => $headers->{$_} } keys %$headers};
```
Which is not good for performance, as for M headers defined in schema we executing `lc` N times (N - total number of headers in HTTP Request). Complexity is `O(M * N)`

A good optimisation is to calculate lc header's hash once and reuse it (which I suggested to do in this PR).

PS: Even just executing to_hash(1) every time is not a good idea, as `sub to_hash` inside Mojo::Headers defined as:
```
sub to_hash {
  my ($self, $multi) = @_;
  return {map { $_ => $self->{headers}{lc $_} } @{$self->names}} if $multi;
  return {map { $_ => $self->header($_) } @{$self->names}};
}
```
ie. it contains a loop inside.

jhthorsen:
Wow! I was sure I had caching of the different parameters, but it went away in 72c33f786fc0dd46aac8633c9534e01dddc105bf

Not sure why I removed it. Anyways: I want to cache all the types, not just `$headers`.

Issue #74, opened 2017-10-04
1.03 doesn't allow remote schemas with ids


wbazant:
Test case:
https://github.com/wbazant/json-validator/commit/27de57c1ae2c5022422919b69963cc4b32b845a0
```
#   Failed test 'remote ref - remote ref valid'
#   at t/acceptance.t line 52.
#          got: '0'
#     expected: '1'
# Invalid schema:
# /id: Does not match uri format. at /Users/wbazant/dev/json-validator/lib/JSON/Validator.pm line 70.
# 	JSON::Validator::load_and_validate_schema('JSON::Validator=HASH(0x7f91a1d0ccf8)', 'HASH(0x7f91a16efa18)') called at t/acceptance.t line 46
# 	eval {...} called at t/acceptance.t line 45
```

jhthorsen:
Is that really a valid ID? Can you point me to the specification?

wbazant:
Sure, that's fair to ask, especially that I didn't check this before submitting the issue! It is:

http://json-schema.org/schema says an id is a uri-reference: http://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3.7

https://tools.ietf.org/html/rfc3986 says an uri-reference can be an URI or a relative-ref (4.1) and (4.2) gives this grammar for relative-ref:
```
relative-ref  = relative-part [ "?" query ] [ "#" fragment ]

      relative-part = "//" authority path-abempty
                    / path-absolute
                    / path-noscheme
                    / path-empty
```

jhthorsen:
I've created a new branch that I'm working on: https://github.com/jhthorsen/json-validator/compare/bug/id-keyword

Could you take a look? It's a work in progress, so you can also wait. (I'll notify you when I make it work)

wbazant:
I had a look! I tried to think abstractly about it, I think the hardest
piece to add is that _load_schema   needs to at least conceptually recurse
on itself and this doesn't (or at least I can see no way in which it does
now)

I'm talking about the first subpart of the problem - making  t/resolve.t work
- parse a schema, possibly with refs, and produce an internal
representation of the schema, without the refs.

Now you have this call stack:
schema
_resolve
_load_schema
...
_load_schema_from_text

together with storing the loaded pieces but it doesn't recurse anywhere,
unless I'm reading it wrong?


I believe the solution has to recurse on itself and I think it has to
involve knowing how to insert the partial results (that got downloaded and
resolved) into the result structure.

[...]

PR #79, opened 2017-10-10
Resolving "$ref" on the fly #65 #74 #75

### LINKED ISSUE #65: json reference validation

jhthorsen:
This pull request makes a major change to the internal representation of the `schema()`: All `$ref` used to be replaced by specification that the ref pointed to. This is now changed. All `$ref` and `id` instances are instead converted to a fully qualified URI and registered internally. Later the `$ref` is looked up when needed - during the validation process.

- Should fix #65, #74 and #75
- Should also fix jhthorsen/mojolicious-plugin-openapi#46
- Removed resolver() attribute - No need for it
- Removed _default_id() - The default ID is just "" (empty string)
- Add RECURSION_LIMIT constant - Meant for internal use
- Might mess up HTML document rendering in mojolicious-plugin-openapi

If you don't have time for the code review, then please install this version and run it with your application's test suite:

```
cpanm https://github.com/jhthorsen/json-validator/archive/bug/id-keyword.tar.gz
```

Considering adding `resolve()` as well: ccefe19


lib/JSON/Validator.pm
```diff
@@ -42,8 +43,7 @@ has cache_paths => sub {
   return \@paths;
 };
 
-has formats  => sub { shift->_build_formats };
-has resolver => sub { \&_resolver };
```
jhthorsen: The "resolver" attribute was marked as EXPERIMENTAL.

lib/JSON/Validator.pm
```diff
@@ -63,33 +63,22 @@ sub coerce {
...
+  $spec = $self->_reset->_resolve($spec);
+  @errors = $clone->schema($args->{schema} || SPECIFICATION_URL)->validate($spec);
+  $self->{schema} = Mojo::JSON::Pointer->new($spec) if !@errors or $args->{want_errors};
+
+  return @errors if $args->{want_errors};    # internal
+  return $self unless @errors;
+  Carp::confess(join "\n", "Invalid schema:", @errors);
 }
 
 sub schema {
```
jhthorsen: The gut of this method is moved to `_resolve()`

lib/JSON/Validator.pm
```diff
@@ -121,58 +110,38 @@ sub _build_formats {
 }
 
 sub _load_schema {
-  my ($self, $url, $parent) = @_;
-  my ($namespace, $scheme) = ("$url", "file");
-  $namespace =~ s|\\|/|g if $^O eq 'MSWin32';
```
jhthorsen: Not sure if I've broken MSWin32 support or improved it.

lib/JSON/Validator.pm
```diff
@@ -196,141 +165,114 @@ sub _load_schema_from_text {
...
 
-  return $tx->res->body;
+  return $self->_load_schema_from_text(\$tx->res->body);
 }
 
-sub _default_id {
-  state $id = 0;
-  return sprintf 'http://local/json-validator/default-id-%s', ++$id;
-}
+sub _reset { delete $_[0]->{refs}; $_[0] }
```
jhthorsen: Not sure if it's needed to delete the whole "refs" key or if it's enough to just delete `$_[0]->{refs}{""}`.

lib/JSON/Validator.pm
```diff
@@ -196,141 +165,114 @@ sub _load_schema_from_text {
...
+    my ($base, $pointer) = split /#/, $ref, 2;
+    my $rel = $self->{refs}{$base // ''} || ($base ? $self->_resolve($base) : $schema);
+    $self->{refs}{$ref} //= Mojo::JSON::Pointer->new($rel)->get($pointer)
+      or Carp::confess(qq[Possibly a typo in schema? Could not find "$ref"]);
+  }
 
   return $schema;
 }
 
-sub _resolver {
```
jhthorsen: Closed over from the `resolver` attribute, so no longer needed.

jberger:
When serving the final version of the spec, would the `$ref`s remain or would they eventually be flattened into the final spec?

jhthorsen:
@jberger: They will remain, because of the recursion issues people have been reporting in. https://github.com/jhthorsen/mojolicious-plugin-openapi/issues/46

jberger:
How would that work with documentation renderers? I don't (currently) serve all the component bits of my spec and indeed I'm splitting it many times. I won't refuse if it fixes lots of other problems but it might be hard on my use case.

jhthorsen:
"if"? Did you see the list of issues that this PR fixes..? 😺

When that is said: I am considering adding a `resolve()` method to JSON::Validator that can return a  perl data structure where all `$ref` instances are replaced. Note however that it's not always possible to convert this into a JSON structure. (Again see the issue I linked to)

I would suppose most document renderers will work though, since they have to work with documents containing $ref. Most servers/sysntes don't flatten, like M::P::OpenAPI does.

lib/JSON/Validator.pm
```diff
@@ -121,58 +110,38 @@ sub _build_formats {
...
-    die "Could not load document from $url: $@";
-  };
-}
+  my $file = $url;
+  $file =~ s!^file://!!;
+  $file =~ s!#$!!;
+  $file = path(split '/', $file);
+  if (-e $file) {
+    $file = $file->realpath;
+    $url = sprintf "file://%s", join "/", @$file;
```
jhthorsen: This is not a good idea. Need to figure out some way to keep relative URLs as-is, so external document renderers download the relative `$ref`.

jhthorsen:
@jberger found a bunch of issues. Need to resolve, or at least figure out, what's going on there before I do any release.

lib/JSON/Validator.pm
```diff
@@ -196,141 +164,133 @@ sub _load_schema_from_text {
...
+
+      push @topics, map { [$topic->{$_}, $base] } grep { $_ ne '$ref' } keys %$topic;
     }
   }
 
-  # Resolve all the "$ref" we found. This will call _resolve_schema() again.
-  $self->resolver->($self, $namespace, \@refs);
+  for (@refs) {
+    my ($topic, $base) = @$_;
+    $topic->{'$ref'} = $self->_resolve_ref($topic->{'$ref'}, $base);
```
jhthorsen: I would much rather replace the `$topic` variable in place with a JSON::Validator::Ref object. Do you know how to do that @jberger or if it's a good idea? If that was done, then I would change the while loop in `_validate()` to:

```
while (UNIVERSAL::isa($schema, 'JSON::Validator::Ref')) {
  Carp::confess("Seems like you have a circular \$ref: $schema") if RECURSION_LIMIT < ++$guard;
  $schema = $schema->schema;
}
```

lib/JSON/Validator/Ref.pm
```diff
@@ -0,0 +1,60 @@
+package JSON::Validator::Ref;
```
jhthorsen: @jberger: This class should fix your issues with the document renderer, since it keeps the original $ref instead of changing it.

lib/JSON/Validator/Ref.pm
```diff
@@ -0,0 +1,60 @@
...
+  q("")    => sub { $_[0]->ref },
+  bool     => sub {1},
+  fallback => 1;
+
+has fqn    => sub { shift->ref };
+has ref    => '';
+has schema => sub { +{} };
+
+# jhthorsen: This cannot return schema() since it might cause circular references
+sub TO_JSON { $_[0]->ref }
```
jhthorsen: I would very much like this to be:

```
sub TO_JSON { {'$ref' => $_[0]->ref} }
```

lib/JSON/Validator.pm
```diff
@@ -816,6 +776,7 @@ sub _is_regex {
 # From Data::Validate::URI
 sub _is_uri {
   return unless $_[0];
+  return 1;    # TODO: I think _is_uri() is broken
```
jhthorsen: I think this URI validation code is broken. I need to look into this... This is the reason for the failing tests now: https://travis-ci.org/jhthorsen/json-validator/jobs/286412268

jhthorsen:
Closing this in favour of #80.

PR #80, opened 2017-10-14
Resolving "$ref" when validating, instead of when parsing the schema

### LINKED ISSUE #65: json reference validation

jhthorsen:
This PR replaces #79: The difference is that instead of manipulating the schema tree, this version will instead register all the refs internally and look them up later.

I did like the idea to replace `{'$ref':'...'}` with a [JSON::Validator::Ref](https://github.com/jhthorsen/json-validator/pull/79/commits/7e36f508ed525751fb9a0ce60aae2cc166c6ff68) object, but it seems to add too many if/else to the code.

If you don't have time for the code review, then please install this version and run it with your application's test suite:

> cpanm https://github.com/jhthorsen/json-validator/archive/bug/id-keyword.tar.gz

- Should fix #65 and #75.
- Should also fix jhthorsen/mojolicious-plugin-openapi/issues#46
- Removed resolver() method - No need for it, and it was marked as EXPERIMENTAL
- Removed _default_id() - The default ID is just "" (emptry string)
- Add RECURSION_LIMIT constant - Meant for internal use
- Might mess up HTML document rendering in mojolicious-plugin-openapi
- Not sure if I've broken MSWin32 support or improved it.
- Considering adding resolve() as well, which will be based on ccefe19

I will merge this on Monday or Tuesday, unless anyone finds a case where this PR does not work.


wbazant:
The code as I pulled is not passing a test that tries to reproduce the problem I had, #65:
https://github.com/wbazant/json-validator/commit/d9db7dfbbbd5c8232ca3e848c576ac2c610f3232
so I think this won't solve the problem for our project yet.

It's also not passing this test:
https://github.com/wbazant/json-validator/commit/46cade5a663fc865563f1e57f285d9dad92a50f6

These don't invalidate the change in any way: it can still be a change for the better. The first case was passing in 0.95, I don't know if the second case was ever supported (it's an artificial case of a valid schema designed to cause problems)

jhthorsen:
I think I've given this enough time now, so merging it.

Thanks to everyone involved!

PR #84, opened 2017-11-24
Update tester script

mohawk2:
Mainly adds `openapi-client` which fails with current version. See that repo for a fix shortly!

### DIFF
--- a/run-all-tests.sh
+++ b/run-all-tests.sh
@@ -30,6 +30,7 @@ elif [ "x$PROJECT" != "x" ]; then
 else
   PROJECT=json-validator t "$@";
   PROJECT=mojolicious-plugin-openapi t "$@";
+  PROJECT=openapi-client t "$@";
 fi
 
 exit $?;
--- a/run-all-tests.sh
+++ b/run-all-tests.sh
@@ -1,9 +1,9 @@
 #!/bin/sh
 # Usage:
-# sh test.sh -j8
-# PROJECT=json-validator sh test.sh -j8
-# HASH_ITERATIONS=10 sh test.sh -v t/plugin-yaml.t
-# PERL_HASH_SEED=8 sh test.sh -v t/plugin-yaml.t
+# sh run-all-tests.sh -j8
+# PROJECT=json-validator sh run-all-tests.sh -j8
+# HASH_ITERATIONS=10 sh run-all-tests.sh -v t/plugin-yaml.t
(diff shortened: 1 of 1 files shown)


run-all-tests.sh
```diff
@@ -1,16 +1,16 @@
...
+# PERL_HASH_SEED=8 sh run-all-tests.sh -v t/plugin-yaml.t
 
 export PERL5LIB="$PWD/lib:$PERL5LIB";
 # export SWAGGER2_DEBUG=1;
 
 t () {
-  echo "\$ cd ../$PROJECT && prove -l $@";
-  cd ../$PROJECT && prove -l $@ || exit $?;
+  echo "\$ cd ../$PROJECT && prove -l $*";
+  cd ../$PROJECT && prove -l "$@" || exit $?;
```
jhthorsen: I don't get why you have quotes around `$@`. What does it fix? I would think that groups all arguments into a single argument when you pass it on to `prove`.

mohawk2:
From the `bash` man page:

```
       @      Expands to the positional parameters, starting from  one.   When
              the  expansion  occurs  within  double  quotes,  each  parameter
              expands to a separate word.  That is, "$@" is equivalent to "$1"
              "$2"  ...   If the double-quoted expansion occurs within a word,
              the expansion of the first parameter is joined with  the  begin‐
              ning  part  of  the original word, and the expansion of the last
              parameter is joined with the last part  of  the  original  word.
              When  there  are no positional parameters, "$@" and $@ expand to
              nothing (i.e., they are removed).
```

run-all-tests.sh
mohawk2: From the `bash` man page:

```
       @      Expands to the positional parameters, starting from  one.   When
              the  expansion  occurs  within  double  quotes,  each  parameter
              expands to a separate word.  That is, "$@" is equivalent to "$1"
              "$2"  ...   If the double-quoted expansion occurs within a word,
              the expansion of the first parameter is joined with  the  begin‐
              ning  part  of  the original word, and the expansion of the last
              parameter is joined with the last part  of  the  original  word.
              When  there  are no positional parameters, "$@" and $@ expand to
              nothing (i.e., they are removed).
```

mohawk2: Tl;dr:

```
# for a script run as `myscript "with    spaces" "hello"`
echo "$@" # preserves spaces
echo $@ # loses spaces, also has wrong number of args
```

jhthorsen:
Cool beans 👍

PR #85, opened 2017-11-25
Make Mojo validator able to get schema from app

mohawk2:
Includes a test to avoid regressions.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -293,6 +293,7 @@ sub _ref_to_schema {
 
 sub _register_schema {
   my ($self, $schema, $fqn) = @_;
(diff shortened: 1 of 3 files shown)


lib/JSON/Validator.pm
```diff
@@ -293,6 +293,7 @@ sub _ref_to_schema {
 
 sub _register_schema {
   my ($self, $schema, $fqn) = @_;
+  $fqn //= 'localhost'; # will be undef if getting schema from app
```
jhthorsen: When is `$fqn` undef? I can't see that in 08291b5.

jhthorsen:
I closed this with 08291b5, since I want JSON::Validator to be able to fetch from local app and not JSON::Validator::OpenAPI::Mojolicious.

PR #87, opened 2017-12-06
Can load $ref from file:/

fyrchik:
Hello.
I have a problem with "$ref" which is defined as "file://...".
I tried to specify "$ref" as an absolute path, in various combinations with "id" -- nothing worked.

If there is another way (not using "file://"), tell me please.
If it is a bug, this PR should fix it.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -219,15 +219,18 @@ sub _load_schema {
     return $self->_load_schema_from_text(\$url), '';
   }
 
-  my $file = $url;
-  $file =~ s!#$!!;
-  $file = path(split '/', $file);
-  if (-e $file) {
-    $file = $file->realpath;
-    warn "[JSON::Validator] Loading schema from file: $file\n" if DEBUG;
-    return $self->_load_schema_from_text(\$file->slurp), $url;
-  }
-  elsif ($file =~ m!^/!) {
+
+  if ($url =~ m!file://(.*)!) {
+    my $file = $1;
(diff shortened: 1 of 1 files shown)


fyrchik:
ok, seems like i have broken some tests :(

jhthorsen:
Why not just use relative path: "../whatever/foo.json" or "bar.json"

jhthorsen:
There's also no failing tests, so it's impossible to know what you are actually trying to fix.

fyrchik:
Oh, yes, it worked.
Thank you!

jhthorsen:
Cool!

fyrchik:
I was trying to trying to understand, why absolute path wath not working, but this PR has broken relative path handling.
Nevertheless, relative paths are OK for me.

jhthorsen:
It's the only way to get it to work consistently. Let's say if you had `"$ref":"/usr/share/whatever/a.json"`, then it would be impossible for a windows developer or someone who wants to test out the spec from a local git repo to make it work.

Also, it would not work if you decided to share the spec online, since then the `$ref` need to resolve to `https://api.example.com/whatever/a.json`.

jhthorsen:
You're welcome to join and ask us questions in the #swagger channel on irc.perl.org.

jhthorsen:
No, it's not. It's relative to where the file doing $ref are.

Here is an example: https://github.com/jhthorsen/json-validator/blob/master/t/spec/with-deep-mixed-ref.json

PR #88, opened 2017-12-24
Added missing prereqs as reported by CPANTS.

manwar:
Hi @jhthorsen 

Please review the PR.

Many Thanks.
Best Regards,
Mohammad S Anwar

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -1,5 +1,7 @@
 # You can install this projct with curl -L http://cpanmin.us | perl - https://github.com/jhthorsen/json-validator/archive/master.tar.gz
 requires "Mojolicious" => "7.15";
+requires "Hash::MultiValue";
+requires "Tie::StdHash";
 
 recommends "Cpanel::JSON::XS"       => "3.02";
 recommends "Data::Validate::Domain" => "0.10";


cpanfile
```diff
@@ -1,5 +1,7 @@
 # You can install this projct with curl -L http://cpanmin.us | perl - https://github.com/jhthorsen/json-validator/archive/master.tar.gz
 requires "Mojolicious" => "7.15";
+requires "Hash::MultiValue";
+requires "Tie::StdHash";
```
jhthorsen: Don't think I need this:

```
corelist Tie::Hash

Data for 2017-09-22
Tie::Hash was first released with perl 5.002
```

cpanfile
```diff
@@ -1,5 +1,7 @@
 # You can install this projct with curl -L http://cpanmin.us | perl - https://github.com/jhthorsen/json-validator/archive/master.tar.gz
 requires "Mojolicious" => "7.15";
+requires "Hash::MultiValue";
```
jhthorsen: This module is optional, in the sense that it is installed by Dancer, if you want to use the Dancer plugin.

jhthorsen:
I can't see how the modules you point out is really missing. Can you elaborate?

manwar:
Hi @jhthorsen 

Here is link to CPANTS that says there are missing prereqs.

https://cpants.cpanauthors.org/release/JHTHORSEN/JSON-Validator-1.08

Best Regards,
Mohammad S Anwar

jhthorsen:
> Though it looks and sounds like quality, higher Kwalitee score doesn't always mean a distribution is more useful for you.

So I don't see how adding those deps increases the quality of the package. Do you actually have issues during installation?

manwar:
I agree with your view on Kwalitee.

However I have come across many distributions that do explicitly list all prereq even if they are in CORE e.g strict, warnings.

Therefore, I thought, it would look nice and clean, when it has all prereqs listed explicitly. There is no harm adding them even if its part of CORE, in my humble opinion.

jhthorsen:
If that is your view, then this PR is incomplete (Missing File::Find) -- I however do not share this view, unless a given version is required.

Sorry.

manwar:
Points well taken and sorry for inconvenience,

jhthorsen:
Not at all: I do appreciate you taking the time and contributing!

Thanks for the interest 👍

perlpunk:
Just a comment:
The prereqs list does not only serve the purpose of listing the modules that need to be installed, but they are also a documentation of what modules are used, without being forced to search the code. This can be used for statistics.

Also, theoretically, modules can be removed from core later; and some distributions (for example fedora/redhat/centos) don't include some core modules in the perl package (although I think this was fixed in recent versions).

Of course, modules that are optional, should not be listed as a prereq.

jhthorsen:
Can't beat two against one ;)

I'll take a PR if it [requires](https://metacpan.org/pod/CPAN::Meta::Spec#requires1) overload, strict, File::Find, File::Spec and File::Temp with the minimum versions. I think I would just find the versions in perl 5.10 and use those.

If you want to do more, you can rewrite the cases where File::Find, File::Spec and File::Temp is used and replace it with Mojo::File.

Also, Hash::MultiValue can be listed as [suggested](https://metacpan.org/pod/CPAN::Meta::Spec#suggests) if you like.

See also https://metacpan.org/pod/distribution/Module-CPANfile/lib/cpanfile.pod#requires,-recommends,-suggests,-conflicts

perlpunk:
@jhthorsen you're right, that's why I said "formally" =)

btw, finding out all prereqs can be done with:
```
scan-perl-prereqs-nqlite . --json
scan-perl-prereqs-nqlite . --cpanfile
```

PR #89, opened 2017-12-27
fix bug coerce numbers vs negative/fraction/exp

umann:
Fix for Bug: when coerce numbers is on, only non-negative integers are validated.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -521,7 +521,7 @@ sub _validate_type_number {
     and $value * 0 == 0)
   {
     return E $path, "Expected $expected - got string."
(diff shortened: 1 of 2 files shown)


jhthorsen:
Why was this closed?

umann:
Because i realized this bug has been fixed, I had a too old version. Sorry.

jhthorsen:
Ah! Cool 👍

mohawk2:
Might be worth reopening and still adding the test that @umann added? I like tests.

jhthorsen:
Both the jv-integer.t and jv-number.t tests check for coercion as well.

PR #91, opened 2018-01-12
Depend on Mojolicious>=7.28

fabzzap:
a8cd35e4a8a8548c30997a5be2d2409fc23449dd added a call to Mojo::File::realpath, that was introduced in Mojolicious 7.28. In practice, this means that now everything depends on Mojolicious>=7.28


jhthorsen:
Thanks! 👍

PR #94, opened 2018-01-21
Correctly identify local fqn when bundling

mohawk2:
Currently, for an OpenAPI schema got from a URL, including a `data:` one, the `fqn` stored in the `JSON::Validator::Ref` contains the URL, which is fine. Unfortunately the `bundle` method special-cases as "local" only `$ref`s that begin with `#`, and does not spot them as "local" if they have the URL.

This patch generalises the "local" check to "starts with the local `fqn`, which might be empty-string".

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -85,7 +85,7 @@ sub bundle {
       my $ref  = ref $from;
 
       if ($ref eq 'HASH' and my $tied = tied %$from) {
-        return $from if $tied->fqn =~ m!^\#!;
+        return $from if $tied->fqn =~ m!^$self->{root_schema_url}\#!;
         my $name = $self->$def_name_cb($tied->fqn);
         push @topics, [$tied->schema, $bundle->{definitions}{$name} = {}];
         tie my %ref, 'JSON::Validator::Ref', $tied->schema, "#/definitions/$name";
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -66,7 +66,7 @@ sub bundle {
   my @topics = ([undef, my $bundle = {}]);
   my ($cloner, $tied);
 
-  $topics[0][0] = $args->{schema} ? $self->_reset->_resolve($args->{schema}) : $self->schema->data;
+  $topics[0][0] = $args->{schema} ? $self->_resolve($args->{schema}) : $self->schema->data;
 
   if ($args->{replace}) {
     $cloner = sub {
@@ -152,7 +152,7 @@ sub get {
 
 sub load_and_validate_schema {
   my ($self, $spec, $args) = @_;
-  $spec = $self->_reset->_resolve($spec);
+  $spec = $self->_resolve($spec);
   my @errors = $self->new(%$self)->schema($args->{schema} || SPECIFICATION_URL)->validate($spec);
   confess join "\n", "Invalid JSON specification $spec:", map {"- $_"} @errors if @errors;
   $self->{schema} = Mojo::JSON::Pointer->new($spec);
@@ -162,7 +162,7 @@ sub load_and_validate_schema {
 sub schema {
   my $self = shift;
   return $self->{schema} unless @_;
-  $self->{schema} = Mojo::JSON::Pointer->new($self->_reset->_resolve(shift));
+  $self->{schema} = Mojo::JSON::Pointer->new($self->_resolve(shift));
   return $self;
 }
 
@@ -328,17 +328,14 @@ sub _report_schema {
   push @{$self->{report}}, [(('  ') x $self->{grouped}) . ('<<<'), $path || '/', $type, D $schema];
 }
 
-sub _reset {
-  delete $_[0]->{schemas}{''};
-  $_[0]->{level} = 0;
-  $_[0];
-}
-
 # _resolve() method is used to convert all "id" into absolute URLs and
 # resolve all the $ref's that we find inside JSON Schema specification.
 sub _resolve {
   my ($self, $schema) = @_;
   my ($id, $resolved, @refs);
+
+  local $self->{level} = $self->{level} || 0;
+  delete $_[0]->{schemas}{''} unless $self->{level};
 
   if (ref $schema eq 'HASH') {
     $id = $schema->{id} // '';
@@ -352,11 +349,13 @@ sub _resolve {
     $id = $schema->{id} if $schema->{id};
   }
 
-  if (!$self->{level}++ and my $id = $schema->{id}) {
-    confess "Root schema cannot have a fragment in the 'id'. ($id)" if $id =~ /\#./;
-    confess "Root schema cannot have a relative 'id'. ($id)" if $id and $id !~ /^\w+:/;
+  unless ($self->{level}) {
+    my $rid = $self->{root_schema_url} = $schema->{id} // $id;
+    confess "Root schema cannot have a fragment in the 'id'. ($rid)" if $rid and $rid =~ /\#./;
+    confess "Root schema cannot have a relative 'id'. ($rid)" if $rid and $rid !~ /^\w+:/;
   }
 
+  $self->{level}++;
   $self->_register_schema($schema, $id);
 
   my @topics = ([$schema, Mojo::URL->new($id)]);
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -350,12 +350,13 @@ sub _resolve {
   }
 
   unless ($self->{level}) {
-    my $rid = $self->{root_schema_url} = $schema->{id} // $id;
+    my $rid = $schema->{id} // $id;
     if ($rid) {
       confess "Root schema cannot have a fragment in the 'id'. ($rid)" if $rid =~ /\#./;
       confess "Root schema cannot have a relative 'id'. ($rid)"
         unless $rid =~ /^\w+:/ or -e $rid or $rid =~ m!^/!;
     }
+    $self->{root_schema_url} = $rid;
   }
 
   $self->{level}++;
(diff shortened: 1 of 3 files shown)


lib/JSON/Validator.pm
```diff
@@ -60,11 +60,18 @@ has ua => sub {
   $ua;
 };
 
+sub _this_fqn {
+  my ($self) = @_;
+  my %schemas = reverse %{ $self->{schemas} || {} };
+  $schemas{ $self->{schema}{data} || '' } // '';
+}
```
jhthorsen: I'm not sure if I'm a huge fan of this method. Could we rather do something like this:

```
my $root_schema = $self->schema->data;
for my $fqn (keys %{$self->{schemas}}) {
  return $fqn if $self->{schemas}{$fqn} eq $root_schema; 
}
return '';
```

lib/JSON/Validator.pm
```diff
@@ -85,7 +92,7 @@ sub bundle {
       my $ref  = ref $from;
 
       if ($ref eq 'HASH' and my $tied = tied %$from) {
-        return $from if $tied->fqn =~ m!^\#!;
+        return $from if $tied->fqn =~ m!^$this_fqn\#!;
```
jhthorsen: What happens if `$this_fqn` is empty string?

t/bundle.t
```diff
@@ -43,4 +43,28 @@ is $validator->get('/name/$ref'), undef,    'get /name/$ref';
 is $validator->schema->get('/name/type'), 'string',             'schema get /name/type';
 is $validator->schema->get('/name/$ref'), '#/definitions/name', 'schema get /name/$ref';
 
+$bundled = $validator->schema('data://main/api.json')->bundle;
+is_deeply [ sort keys %{ $bundled->{definitions} } ], [ 'objtype' ], 'no dup definitions';
```
jhthorsen: Could you add this test to the other schemas above as well? I also think we should write a test where we load a schema from disk, with `$ref` to another file on disk. Like this: https://github.com/jhthorsen/json-validator/blob/master/t/deep-mixed-ref.t

lib/JSON/Validator.pm
mohawk2: That's the current behaviour. See the line it's replacing?

lib/JSON/Validator.pm
mohawk2: I did consider a for-loop, or perhaps a `grep`. But then I thought that would be a slightly-harder to understand, and slower, way to achieve what a hash is designed for. Do you really insist on a for-loop?

t/bundle.t
mohawk2: The other cases in `t/bundle.t` by definition (ha!) already work, see especially the line above here - the `$ref` is proved to be correct. The first bit of the file also can't behave differently because the `$fqn` will be empty-string.

I'll add similar testing to `t/deep-mixed-ref.t` right now.

mohawk2: Done, see new commit.

lib/JSON/Validator.pm
jhthorsen: I must have been sleeping. Thank you.

lib/JSON/Validator.pm
jhthorsen: How about this then: (Look for `root_schema_id`)

```diff
diff --git a/lib/JSON/Validator.pm b/lib/JSON/Validator.pm
index 6cc4ccd..e24ae49 100644
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -66,7 +66,7 @@ sub bundle {
   my @topics = ([undef, my $bundle = {}]);
   my ($cloner, $tied);

-  $topics[0][0] = $args->{schema} ? $self->_reset->_resolve($args->{schema}) : $self->schema->data;
+  $topics[0][0] = $args->{schema} ? $self->_resolve($args->{schema}) : $self->schema->data;

   if ($args->{replace}) {
     $cloner = sub {
@@ -152,7 +152,7 @@ sub get {

 sub load_and_validate_schema {
   my ($self, $spec, $args) = @_;
-  $spec = $self->_reset->_resolve($spec);
+  $spec = $self->_resolve($spec);
   my @errors = $self->new(%$self)->schema($args->{schema} || SPECIFICATION_URL)->validate($spec);
   confess join "\n", "Invalid JSON specification $spec:", map {"- $_"} @errors if @errors;
   $self->{schema} = Mojo::JSON::Pointer->new($spec);
@@ -162,7 +162,7 @@ sub load_and_validate_schema {
 sub schema {
   my $self = shift;
   return $self->{schema} unless @_;
-  $self->{schema} = Mojo::JSON::Pointer->new($self->_reset->_resolve(shift));
+  $self->{schema} = Mojo::JSON::Pointer->new($self->_resolve(shift));
   return $self;
 }

@@ -328,18 +328,15 @@ sub _report_schema {
   push @{$self->{report}}, [(('  ') x $self->{grouped}) . ('<<<'), $path || '/', $type, D $schema];
 }

-sub _reset {
-  delete $_[0]->{schemas}{''};
-  $_[0]->{level} = 0;
-  $_[0];
-}
-
 # _resolve() method is used to convert all "id" into absolute URLs and
 # resolve all the $ref's that we find inside JSON Schema specification.
 sub _resolve {
   my ($self, $schema) = @_;
   my ($id, $resolved, @refs);

+  local $self->{level} = $self->{level} || 0;
+  delete $_[0]->{schemas}{''} unless $self->{level};
+
   if (ref $schema eq 'HASH') {
     $id = $schema->{id} // '';
     return $resolved if $resolved = $self->{schemas}{$id};
@@ -352,11 +349,13 @@ sub _resolve {
     $id = $schema->{id} if $schema->{id};
   }

-  if (!$self->{level}++ and my $id = $schema->{id}) {
-    confess "Root schema cannot have a fragment in the 'id'. ($id)" if $id =~ /\#./;
-    confess "Root schema cannot have a relative 'id'. ($id)" if $id and $id !~ /^\w+:/;
+  unless ($self->{level}) {
+    my $rid = $self->{root_schema_id} = $schema->{id} // '';
+    confess "Root schema cannot have a fragment in the 'id'. ($rid)" if $rid and $rid =~ /\#./;
+    confess "Root schema cannot have a relative 'id'. ($rid)" if $rid and $rid !~ /^\w+:/;
   }

+  $self->{level}++;
   $self->_register_schema($schema, $id);

   my @topics = ([$schema, Mojo::URL->new($id)]);
```

I want to apply this patch either way, to avoid the whole `_reset()` method, but you're welcome to use this in this PR.

t/bundle.t
```diff
@@ -43,4 +44,38 @@ is $validator->get('/name/$ref'), undef,    'get /name/$ref';
...
+
+my $file
+  = File::Spec->catfile(File::Basename::dirname(__FILE__), 'spec', 'with-deep-mixed-ref.json');
```
jhthorsen: I'm gonna change to `Mojo::File` all over the place.

t/bundle.t
```diff
@@ -43,4 +44,38 @@ is $validator->get('/name/$ref'), undef,    'get /name/$ref';
...
+$bundled = $validator->schema($file)->bundle;
+is_deeply [ sort keys %{ $bundled->{definitions} } ], [qw(
+  _t_definitions_age_json
+  _t_definitions_unit_json
+  _t_definitions_weight_json
+  height
+)], 'right definitions in disk spec';
```
jhthorsen: 👍

lib/JSON/Validator.pm
mohawk2: The above also needed this, which I've applied on the branch of this PR together with your patch:

```diff
diff --git a/lib/JSON/Validator.pm b/lib/JSON/Validator.pm
index daf30f1..ace94d3 100644
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -351,8 +351,11 @@ sub _resolve {
 
   unless ($self->{level}) {
     my $rid = $self->{root_schema_url} = $schema->{id} // $id;
-    confess "Root schema cannot have a fragment in the 'id'. ($rid)" if $rid and $rid =~ /\#./;
-    confess "Root schema cannot have a relative 'id'. ($rid)" if $rid and $rid !~ /^\w+:/;
+    if ($rid) {
+      confess "Root schema cannot have a fragment in the 'id'. ($rid)" if $rid =~ /\#./;
+      confess "Root schema cannot have a relative 'id'. ($rid)"
+        unless $rid =~ /^\w+:/ or -e $rid or $rid =~ m!^/!;
+    }
   }
 
   $self->{level}++;
```

mohawk2:
I applied the `_reset`-removing patch, together with a needed fix and a tidy-up, and used the `root_schema_url` instead. I hope this now works for you?

lib/JSON/Validator.pm
```diff
@@ -352,11 +349,17 @@ sub _resolve {
...
-    confess "Root schema cannot have a fragment in the 'id'. ($id)" if $id =~ /\#./;
-    confess "Root schema cannot have a relative 'id'. ($id)" if $id and $id !~ /^\w+:/;
+  unless ($self->{level}) {
+    my $rid = $schema->{id} // $id;
+    if ($rid) {
+      confess "Root schema cannot have a fragment in the 'id'. ($rid)" if $rid =~ /\#./;
+      confess "Root schema cannot have a relative 'id'. ($rid)"
+        unless $rid =~ /^\w+:/ or -e $rid or $rid =~ m!^/!;
+    }
+    $self->{root_schema_url} = $rid;
```
jhthorsen: Maybe this could be moved up?

```
my $rid = $self->{root_schema_url} = $schema->{id} // $id;
```

jhthorsen:
Awesome work @mohawk2! 👍 

Thanks for applying my `_reset()` patch as well.

PR #96, opened 2018-01-27
Fix multitype enum validation inside an allOf

ugexe:
Fixes regression starting in v2.00 where an enum containing a type constraint of `["string","null"]` would not accept return values ( from Mojolicious::Plugin::OpenAPI ) of `undef` as valid if inside an allOf.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -558,6 +558,7 @@ sub _validate_type_enum {
   my $m    = S $data;
 
   for my $i (@$enum) {
+    return if $m eq S(undef) && S($i) eq S('null');
     return if $m eq S $i;
   }
 


ugexe:
There is probably a better fix than this PR, but hopefully this helps demonstrate its goal. Maybe the solution even belongs in Mojolicious::Plugin::OpenAPI?

Note I have not had time to construct a test, although this PR does pass current tests. Below is an example of the OpenAPI spec which allows a return value of `attributes : 'null'` in 1.08 but not in 2.00 or 2.01.

```
{
  "definitions": {
    "stipulation" : {
      "title" : "Data",
      "type" : "object",
      "allOf" : [
        {
          "$ref" : "#/definitions/metadata"
        },
        {
          "type" : "object",
          "properties" : {
            "name" : {
              "type" : "string"
            }
          }
        }
      ]
    },
    "metadata" : {
      "title" : "Metadata",
      "type" : "object",
      "properties" : {
        "attributes" : {
          "type" : ["string","null"],
          "enum" : ["foo", "bar","null"]
        }
      }
    }
  }
}
```

jhthorsen:
Are you sure `"enum" : ["foo", "bar","null"]` is right? I thought you had to have `"enum" : ["foo", "bar",null]`

ugexe:
Having both enum and type has ( or was ) the only way to get an enumeration to accept both string values *or* null. Additionally it had to be a quoted null for the app to work properly ( writen approximately 1.5 years ago ), at least up to JV@1.08

jhthorsen:
But `"null"` isn't the same as `null`, don't you agree..?

ugexe:
https://github.com/jhthorsen/mojolicious-plugin-openapi/blob/5a0f0a2538b79e752f6d8db2b4ed55b7c0aedf31/t/empty-string.t#L43

`"schema": {"type": ["null", "string"]}`

jhthorsen:
Type is not the same as enum.

ugexe:
So I just applied your advice and removed the `type` and changed `"null"` to `null` inside the enum, and the app works with the newer JSON::Validator.

I don't remember *why* we had to do that odd type/enum combo originally ( I remember wondering why we had to do it that way, but without time to investigate ). If I re-encounter the situation I will revisit this with additional info, but hopefully the cause was simply inadvertently fixed by other updates somewhere.

jhthorsen:
Glad it worked out for you 👍

ugexe:
So the reason we had quoted nulls and the extra type info was because swagger-ui 2.x does not work otherwise. It *does* work with swagger-ui 3.x though, so all is good.

PR #97, opened 2018-02-19
Catch duplicate "body" parameters

mohawk2:
This should help a bit with #15.


jhthorsen:
It seems like an awful lot of code to catch something that is obvious.

I will think about it, but will probably not reach any conclusion until next week.

jhthorsen:
I don't think it's worth it, but I'll keep it open a bit longer for others to comment. Ask in the IRC channel for support and I will for sure reconsider.

mohawk2:
35 lines of code is "an awful lot" to catch something that's only obvious if you know to watch out for it?

jhthorsen:
I've implemented the same logic in b09ff5a now, so I'm closing this issue.

Thanks for your contribution!

mohawk2:
@jhthorsen Does this belatedly mean #15 can also be closed?

PR #105, opened 2018-03-31
Test problem seen when directory name has RFC 3986 reserved chars

mohawk2:
The "definitions" file has a space here. I am using this to reproduce the problem that shows up (thanks to the call to `Mojo::File::realpath`) when any character that `Mojo::URL` percent-encodes is present anywhere in the full directory path. This is because the encoded version is then used as if it is a viable filesystem location, but it is not due to the encoding.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -76,7 +76,7 @@ sub bundle {
 
       if ($ref eq 'HASH' and my $tied = tied %$from) {
         my $ref_name = $tied->fqn;
-        return $from if $ref_name =~ m!^$self->{root_schema_url}\#!;
+        return $from if $ref_name =~ m!^\Q$self->{root_schema_url}\E\#!;
 
         if (-e $ref_name) {
           $ref_name = sprintf '%s-%s', substr(sha1_sum($ref_name), 0, 10),
@@ -250,9 +250,9 @@ sub _load_schema {
   if (-e $file) {
     $file = $file->realpath;
     warn "[JSON::Validator] Loading schema from file: $file\n" if DEBUG;
-    return $self->_load_schema_from_text(\$file->slurp), CASE_TOLERANT ? lc $file : $file;
+    return $self->_load_schema_from_text(\$file->slurp), CASE_TOLERANT ? path(lc $file) : $file;
   }
-  elsif ($file =~ m!^/!) {
+  elsif ($url =~ m!^/!) {
     warn "[JSON::Validator] Loading schema from URL $url\n" if DEBUG;
     return $self->_load_schema_from_url(Mojo::URL->new($url)->fragment(undef)), "$url";
   }
@@ -394,7 +394,7 @@ sub _resolve {
   $self->{level}++;
   $self->_register_schema($schema, $id);
 
-  my @topics = ([$schema, Mojo::URL->new($id)]);
+  my @topics = ([$schema, blessed($id)//'' eq 'Mojo::File' ? $id : Mojo::URL->new($id)]);
   while (@topics) {
     my ($topic, $base) = @{shift @topics};
 
@@ -423,6 +423,14 @@ sub _resolve {
 sub _absolutise_location {
   my ($location, $base) = @_;
   my $location_as_url = Mojo::URL->new($location);
+  return $location_as_url if $location_as_url->is_abs;
+  # definitely relative now
+  my $ref = blessed $base
+    or die "[JSON::Validator] \$ref location base was non-object";
+  if ($ref eq 'Mojo::File') {
+    return $base if !length $location;
+    return $base->sibling(split '/', $location)->realpath;
+  }
   return $location_as_url->to_abs($base);
 }
 
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -437,7 +437,7 @@ sub _resolve_ref {
   return if tied %$topic;
 
   my $other = $topic;
-  my ($base, $fqn, $pointer, $ref, @guard);
+  my ($location, $fqn, $pointer, $ref, @guard);
 
   while (1) {
     $ref = $other->{'$ref'};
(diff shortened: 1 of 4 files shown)


mohawk2:
This is a test that demonstrates the problem when a file (or a directory that file lives in) has e.g. a space in it. I wanted to ensure you were interested before I worked on a fix for it.

The currently-released code does not work on my "Perl in space" setup, which you can see in the cpantesters report: `t/bundle.t` and `t/load-from-app.t` fail. If you're interested in a fix, let me know.

mohawk2:
One problem is in `Validator.pm` line 79:

```perl
        return $from if $ref_name =~ m!^$self->{root_schema_url}\#!;
```

needs to become:

```perl
        return $from if $ref_name =~ m!^\Q$self->{root_schema_url}\E\#!;
```

The `load-from-app.t` failure is because `/api` gets turned into `\api` on Windows, because the current code is treating a URI as though it's a file location.

jhthorsen:
Could you push a fix as well, and squash the commits into one?

mohawk2:
@jhthorsen This is now ready for review, and actually passes its tests on Windows too!

I left it in two commits because the first one seemed genuinely separate though preparatory for the main change. Let me know if you need any further changes.

lib/JSON/Validator.pm
```diff
@@ -250,9 +250,9 @@ sub _load_schema {
   if (-e $file) {
     $file = $file->realpath;
     warn "[JSON::Validator] Loading schema from file: $file\n" if DEBUG;
-    return $self->_load_schema_from_text(\$file->slurp), CASE_TOLERANT ? lc $file : $file;
+    return $self->_load_schema_from_text(\$file->slurp), CASE_TOLERANT ? path(lc $file) : $file;
```
jhthorsen: Why path on `lc $file` but not on `$file`?

lib/JSON/Validator.pm
```diff
@@ -394,7 +394,7 @@ sub _resolve {
   $self->{level}++;
   $self->_register_schema($schema, $id);
 
-  my @topics = ([$schema, Mojo::URL->new($id)]);
+  my @topics = ([$schema, blessed($id)//'' eq 'Mojo::File' ? $id : Mojo::URL->new($id)]);
```
jhthorsen: Could you do `UNIVERSAL::isa($id, 'Mojo::File')` instead? Also, this does not look like it's been run through perltidy.

lib/JSON/Validator.pm
```diff
@@ -420,6 +420,20 @@ sub _resolve {
   return $schema;
 }
 
+sub _absolutise_location {
```
jhthorsen: Could you call this `_location_to_abs()` instead?

lib/JSON/Validator.pm
```diff
@@ -420,6 +420,20 @@ sub _resolve {
...
+  my ($location, $base) = @_;
+  my $location_as_url = Mojo::URL->new($location);
+  return $location_as_url if $location_as_url->is_abs;
+  # definitely relative now
+  my $ref = blessed $base
+    or die "[JSON::Validator] \$ref location base was non-object";
+  if ($ref eq 'Mojo::File') {
```
jhthorsen: Instead of using $ref, then just do `$base->isa('Mojo::File')` instead. And I don't want the `die(...)` above. This is not user input, so a unit test should be enough.

lib/JSON/Validator.pm
```diff
@@ -432,14 +446,12 @@ sub _resolve_ref {
...
-    $fqn = Mojo::URL->new($ref =~ m!^/! ? "#$ref" : $ref);
-    $fqn = $fqn->to_abs($url) unless $fqn->is_abs;
-    $url = $fqn;
-    $fqn = $fqn->to_string;
-    $fqn =~ s!(.)#$!$1!;
-    $fqn =~ s!#(.+)!{'#' . url_unescape $1}!e;
-
+    $fqn = $ref =~ m!^/! ? "#$ref" : $ref;
     ($base, $pointer) = split /#/, $fqn, 2;
+    $url = $base = _absolutise_location($base, $url);
```
jhthorsen: Not related to this PR, but I find my variables `$url` and `$base` confusing. You're free to rename them if you have a better suggestion.

jhthorsen:
I think the PR looks very promising! Thanks for the work 👍

lib/JSON/Validator.pm
mohawk2: 2 lines above, it's visibly a `Mojo::File`. If you `lc` that, it stringifies it, and `path` then turns it back to a `Mojo::File`. It needs to be that, because if it's a string, it gets made into a `Mojo::URL`, which causes problems.

lib/JSON/Validator.pm
mohawk2: I've renamed `$base` to `$location` in line with your suggested rename of the absolute-ising function.

lib/JSON/Validator.pm
jhthorsen: Ah! Thank you 👍

lib/JSON/Validator.pm
```diff
@@ -420,31 +420,41 @@ sub _resolve {
...
+sub _location_to_abs {
+  my ($location, $base) = @_;
+  my $location_as_url = Mojo::URL->new($location);
+  return $location_as_url if $location_as_url->is_abs;
+  # definitely relative now
+  if ($base->isa('Mojo::File')) {
+    return $base if !length $location;
+    return $base->sibling(split '/', $location)->realpath;
+  }
+  return $location_as_url->to_abs($base);
```
jhthorsen: This is a minor thing that I can do myself, but I guess you want credit for your own lines of code. How about something like this:

```
sub _location_to_abs {
  my ($location, $base) = @_;
  my $location_as_url = Mojo::URL->new($location);
  return $location_as_url if $location_as_url->is_abs;    # Already absolute
  return $location_as_url->to_abs($base) unless $base->isa('Mojo::File');    # Mojo::URL to absolute
  return $base unless length $location;                                      # No sibling file to locate
  return $base->sibling(split '/', $location)->realpath;                     # Find sibling file relative to $base
}
```

jhthorsen: (not sure if the comments are needed)

mohawk2: I think your version is better and am pleased if you update it to that. I think the comments do add something.

jhthorsen:
Thank you and sorry for the late review 👍

PR #107, opened 2018-04-03
Turn OpenAPI v2 tests from code into data-driven

mohawk2:
This is not yet ready for merging and is created for early comment.

As is visible, the concept is to move all the code tests for generic OpenAPI v2 behaviour (as opposed to the Perl-specific stuff like handling of booleans) into data, in the same spirit as, and based on the format of, the draft-04 tests for JSON-Schema.


jhthorsen:
I don't see what kind of enhancement this change adds.

mohawk2:
If we have data-driven tests for OpenAPI v2, they can be translated into v3 tests. This is something the OAI folks have expressed interest in.

I based this branch off my other failing-test branch with the intention to capture that behaviour in the tests. I have now rethought that and will shortly rebase this off master minus that, so the CI will be passing again.

mohawk2:
So to be clear, you're saying you're not at all interested in making the tests data-driven? I confess that I'm surprised. Or could you expand on the "as it stands" a bit so I understand you better?

jhthorsen:
It's not that I don't want data-driven tests. As you already know, that's what drives acceptance.t. 

1. It looks like it contains the commits from #105
2. I think I would like one file per test case, like `t/openapi-tests/v2/validation-of-readOnly.json` and `t/openapi-tests/v2/format-byte.json`
3. I don't want four spaces in the JSON files (I want two)
4. Maybe less verbose test files. (Same content, but less lines..? Not sure)
5. I want the shared code between acceptance.t to be put into `t/Helper.pm`
6. Maybe rename `t/acceptance.t` into `t/draft4.t`

mohawk2:
All good points, I will work on those. Do you want to re-open this PR?

PR #110, opened 2018-04-16
Fix strict for object issues

pamasse:
In Validator.pm line 763, you check if "additionalProperties" exists in $schema but in Joi.pm you create a "additionalItems" key when "strict" is used.
As the additionalProperties key is never created, the strict option never works.

### DIFF
--- a/lib/JSON/Validator/Joi.pm
+++ b/lib/JSON/Validator/Joi.pm
@@ -98,7 +98,7 @@ sub _compile_object {
   my $self = shift;
   my $json = {type => $self->type};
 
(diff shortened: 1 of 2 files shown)


mohawk2:
@pamasse Great catch. Could you add a test that fails without this patch, to avoid future regressions?

pamasse:
Yes of course, i will do that tomorrow :)

pamasse:
@mohawk2 Do you think that is possible to merge this pull request this week and repackage your lib in CPAN ?
Thanks a lot for your work :)

jhthorsen:
@pamasse: Version 2.07 should be available soon on CPAN.

PR #112, opened 2018-05-04
protect against XXE DOS attacks on external refs

leejo:
there's a new(ish) kind of DOS attack in the OWASP top ten: XXE[1],
where referring to an external entity that never returns can lead to DOS
attacks when parsing XML documents as they try to load those
never-ending external entities.

although they refer to XML there, it's pretty clear this can also affect
JSON-Schema and here's a test case that shows this. the fix has been
made in Validator.pm to prevent loading of files from /dev - this is a
pretty crude fix, which perhaps should be replaced with something more
generic like a timeout?

note there are more interesting attacks if we reference /etc/passwd:

    Can't use string ("nobody:*:-2:-2:Unprivileged User"...) as a
        HASH ref while "strict refs" in use at
        lib/JSON/Validator.pm line 269.

this is probably to discuss, but for now here's a trivial test case +
crude fix that show the POC in action

[1] https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -243,6 +243,8 @@ sub _load_schema {
   $file = path(split '/', $file);
   if (-e $file) {
     $file = $file->realpath;
+    confess "Unable to load schema due to potential XXE attack vector: $file"
+      if $file =~ m!^/dev/!;
     warn "[JSON::Validator] Loading schema from file: $file\n" if DEBUG;
     return $self->_load_schema_from_text(\$file->slurp), CASE_TOLERANT ? lc $file : $file;
   }
--- a/t/load-xxe.t
+++ b/t/load-xxe.t
@@ -1,0 +1,32 @@
+use lib '.';
+use t::Helper;
+use Test::More;
+use Mojo::Path;
+
+my $path = Mojo::File->new('/dev/random')
+  ->to_rel( Mojo::File->new->path );
+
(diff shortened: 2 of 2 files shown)


jhthorsen:
JSON::Validator has built in protection from this in the sense that you can easily bundle all external refs with your application or in a shared directory.

When that is said, I don't mind this fix. The only thing I would like feedback on is cross platform, and is "/dev/" the only path? What about "/proc"?

Last, how is this even possible? Shouldn't "/dev" be relative to the base URL of the schema loaded? Meaning if you load from https://api.example.com/myschema.json, then "/dev/hda1" should resolve to https://api.example.com/dev/hda1 and not file:///dev/hda1

leejo:
> Last, how is this even possible? Shouldn't "/dev" be relative to the base URL of the schema loaded? Meaning if you load from https://api.example.com/myschema.json, then "/dev/hda1" should resolve to https://api.example.com/dev/hda1 and not file:///dev/hda1

It's working with files here, at least the test is. This is an edge case, but I guess if you're consuming JSON-Schema from an external source then that's where the exploit can come from.

jhthorsen:
That's not the test case is about. The test case is loaded from the local filesystem. I don't think you can make a remote schema load from /dev/random from the local system, since it will be resolved against the base URL of the remote schema. Therefor I will close this PR.

PR #117, opened 2018-07-18
Make validate() return a list of errors when failing "allOf", closes #116

### LINKED ISSUE #116: Could validate() return a list of errors when failing "allOf"?
Usually I get a list of JSON::Validator::Error with each path and message.

Shouldn't it be the same with "allOf" ?

In the following example I get the two errors on the same message and no reference to which property name.

````
#!/usr/bin/env perl

use JSON::Validator;
use JSON;
use Data::Printer;

my $validator = JSON::Validator->new;

my $openapi = '{
    "Service" : {
        "type": "object",
        "properties": {
            "prop1": { "type": "string" },
            "prop2": { "type": "string" }
        }
    },
    "allOfThem" : {
        "allOf": [
            { "$ref": "#/Service" },
            {
                "properties": {
                    "prop3": { "type": "string" },
                    "prop4": { "type": "string" }
                },
                "required": [ "prop3" ]
            }
        ]
    }
}';

$validator = $validator->schema($openapi);

my $sub_schema = $validator->get('/allOfThem');

my $data_json = <<EOT
{
  "prop1": "foo",
  "prop4": 666
}
EOT
;

my $data = from_json($data_json);

my @errors = $validator->validate( $data, $sub_schema);

p @errors;

```

```
[
    [0] JSON::Validator::Error  {
        Parents       Mojo::Base
        public methods (5) : has, message, new, path, TO_JSON
        private methods (1) : __ANON__
        internals: {
            message   "allOf failed: (Missing property.. Expected string - got number.)",
            path      "/"
        }
    }
]

```

This happens with 2.08

Thank you!

biafra:
This is a point of concept for getting each error for each proper parameter in complex structures. You just turn on an option to get all the errors in an individual JSON::Validator::Error

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -1344,6 +1344,12 @@ Default is to use the value from the L</schema> attribute.
 
 =back
 
+=head2 errors_conf
+
+  $self = $self->errors_conf( complex_as_list => 1 );
+
+Set how to return the errors of C<validate> if it's a complex structure (all/any/oneOf and nested). The default is just one JSON::Validator::Error with C<message> concatenated.
+
 =head2 coerce
 
   $self = $self->coerce(booleans => 1, numbers => 1, strings => 1);
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -114,6 +114,14 @@ sub bundle {
   return $bundle;
 }
 
+# $self->errors_conf( complex_as_list => 1 );
+sub errors_conf {
+  my $self = shift;
+  return $self->{errors_conf} ||= {} unless @_;
+  $self->{errors_conf} = ref $_[0] ? {%{$_[0]}} : {@_};
+  $self;
+}
+
 sub coerce {
   my $self = shift;
   return $self->{coerce} ||= {} unless @_;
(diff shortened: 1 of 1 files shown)


lib/JSON/Validator.pm
```diff
@@ -114,6 +114,14 @@ sub bundle {
   return $bundle;
 }
 
+# $self->errors_conf( complex_as_list => 1 );
+sub errors_conf {
```
jhthorsen: Why is this a method and not a simple attribute like `has errors_conf => sub { +{} };`?

lib/JSON/Validator.pm
```diff
@@ -529,7 +537,15 @@ sub _validate_all_of {
   my $expected = join ' or ', _uniq(@expected);
   return E $path, "allOf failed: Expected $expected, not $type."
     if $expected and @errors + @expected == @$rules;
-  return E $path, sprintf 'allOf failed: %s', _merge_errors(@errors) if @errors;
+
+  if (@errors) {
+    if ( $self->{errors_conf}->{complex_as_list} ) {
+        return _flatten(@errors);
```
jhthorsen: How will you know that it was an allOf error when you just return a list of errors?

jhthorsen:
This PR is a proposed solution to #116.

lib/JSON/Validator.pm
biafra: The idea was just to keep the door open for other possible options like in "coerce".

lib/JSON/Validator.pm
biafra: We don't. But is it important ? This is really a question for me to understand sorry. Isn't all the "...Of" just a way to group "fields" or "objects" ?

On the example of issue #116 I thought I could reference a property by '/allOfThem/prop3". Now I'm seeing it should be 'allOfThem/allOf/prop3' ? That's awfull :-(

biafra:
This was more to test a way to report better error validation than a complete solution, I'll digest your ideas. Thank you!

lib/JSON/Validator.pm
jhthorsen: coerce() is a method for back compat. No need for it here when it's a new attribute.

lib/JSON/Validator.pm
jhthorsen: I think we need a complex test case to see how it plays out in real life.

jhthorsen:
Closing this based on the comment above and inactivity.

PR #121, opened 2018-10-20
Will flatten allOf, anyOf and oneOf errors, closes #15 #116

### LINKED ISSUE #15: Errors in anyOf validation. Do not understand the error messages

jhthorsen:
I think I was very wrong in https://github.com/jhthorsen/json-validator/issues/116#issuecomment-405523820 when I suggested adding "children" to JSON::Validator::Error, since it would make it a lot harder for the end user to figure out what was wrong.

I also think I've been previously wrong when I said that allOf, anyOf and oneOf should only return one error. I can't at this point not recall why I thought that was the only way.

This PR suggests that allOf, anyOf and oneOf will return a list of errors, one for each unique failure that happens. In addition, each failure message will be prefixed with a relative path to the validation object that made this fail.

Please review the changes in the test suite, since I think that will make it more obvious what is going on.

You, @biafra, @elowy01, @KES777, might be interested into having a look as well.

Does this change require a major version number bump?

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -527,20 +527,24 @@ sub _validate_all_of {
   $self->_report_schema($path, 'allOf', $rules) if REPORT;
   $self->{grouped}++;
 
+  my $i = 0;
   for my $rule (@$rules) {
     my @e = $self->_validate($data, $path, $rule) or next;
     my $schema_type = _guess_schema_type($rule);
-    push @errors, [@e] and next if !$schema_type or $schema_type eq $type;
+    push @errors, [$i, @e] and next if !$schema_type or $schema_type eq $type;
     push @expected, $schema_type;
+  }
+  continue {
+    $i++;
   }
 
   $self->{grouped}--;
 
   $self->_report_errors($path, 'allOf', \@errors) if REPORT;
   my $expected = join ' or ', _uniq(@expected);
-  return E $path, "allOf failed: Expected $expected, not $type."
+  return E $path, "/allOf Expected $expected, not $type."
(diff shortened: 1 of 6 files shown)


KES777:
I will not agree with you. The error message is `[ $path, $message ]`
[here](https://github.com/jhthorsen/json-validator/pull/121/commits/926da4cc6f3af0a2a6933124648e6d7a5a6181b1#diff-adc6881507836d26fecdde466d40eea1R567)

You put just only `path` in front of array of errors. And no message for current path.

Yes, you put the errors for children schema of 'anyOf', but message for current schema is lost.

I would suggest next instead of just `$i`:
```
E($i,'All rules should be matched')
```

jhthorsen:
What are you talking about @KES777? The `$path` is the path to the input document, while `$message` contains what went wrong, but also which part of the schema that caused it to go wrong.

Side note: In the cases where allOf, anyOf and oneOf is not involved, the error message will just have "what went wrong", since the path is the same for both input and output.

Please comment on the changes in the tests instead.

biafra:
@jhthorsen After my first tests this is really what I asked for! (just used AllOf)

PR #122, opened 2018-10-29
Revert Mojolicious version dependency bump

ilmari:
The change to cope with the changed Mojo::URL->query() DWIM semantics
in Mojolicious 8.00 switched to using an explit ->merge() call, which
has worked the same way forever, so there's no need to bump the
required version of Mojolicious.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -9,7 +9,7 @@ my %WriteMakefileArgs = (
   EXE_FILES      => [qw()],
   BUILD_REQUIRES => {},
   TEST_REQUIRES  => {'Test::More' => '1.30'},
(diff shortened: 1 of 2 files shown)


jhthorsen:
Could we bump the version a bit?

ilmari:
I considered that, but didn't see any relevant changes to Mojo::URL or Mojo::Parameters, so I just reverted to the previous version. If there are other relevant fixes in 7.x, feel free to bump to whatever version has them (we're on 7.94 at work)

ilmari:
@jhthorsen as requested on IRC: *prodprod* ;-P

PR #123, opened 2018-11-12
Improved validation of OpenAPI date format

jlcooper:
Currently the OpenAPI `date` format validates against the regular expression pattern of `^(\d+)-(\d+)-(\d+)$` which allows a lot of common mistakes through (`31-01-1970`, `01-31-2014`, `31-13-1970`, etc).

I've added some additional tests to for the `date` format and updated the `_is_date` function in `JSON::Validator::OpenAPI` to have stricter validation rules based on RFC3339 (i.e. `YYYY-MM-DD`).


jhthorsen:
Thanks for the PR! I decided to go in a bit different direction, but your failing tests were very helpful 👍 

2.16 should be available on CPAN soon.

PR #128, opened 2018-12-09
Tidy up the codebase

jhthorsen:


lib/JSON/Validator.pm
```diff
@@ -315,6 +311,15 @@ sub _load_schema_from_url {
...
 }
 
+sub _matcher {
+  my ($format, $module, $method) = @_;
+  my $e = eval "require $module;1" ? undef : $@;
+  return sub { warn "$module is not available: $e"; return undef }
+    if $e;
+  my $m = $module->can($method);
+  return sub { $m->($_[0]) ? undef : "Does not match $format format." };
+}
```
kraih: Wouldn't this work with [Mojo::Loader::load_class](https://mojolicious.org/perldoc/Mojo/Loader#load_class)?

jhthorsen: I don’t think those modules have a new() method, so I wasn’t sure if it was awkward with load_class

kraih: They both appear to have new() methods.

PR #130, opened 2018-12-12
Add support for translating error messages

jhthorsen:
To be able to merge this PR, I need at least someone to look over the phrases documented in JSON::Validator::Error.

I do however think it's a good idea for the format validators to return an array-ref, instead of a plain string.

Overview of PR:

* The changes in JSON::Validator is just to be compatible with the new JSON::Validator::Error API.
* The changes in JSON::Validator::Error is what is interesting.
* The changes in the tests should also be looked over.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Revision history for perl distribution JSON-Validator
 
 3.00 Not Released
+ - Add support for translating error messages
  - Fix cases where input was not coerced
  - Changed how format validation is done
  - Changed Joi to always coerce values
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -317,7 +317,7 @@ sub _matcher {
   return sub { warn "$module is not available: $e"; return undef }
     if $e;
   my $m = $module->can($method);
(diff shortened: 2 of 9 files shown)


beyondcreed:
Phrases look ok to me.

beyondcreed:
However I think what the original idea was, that client code would translate directly, not the backend. Most of our use cases are SPA's or PWA's that read some data via API, interpret error codes and then select appropriate error on their own by consulting local lookup table. I guess with this approach you are thinking more the translation should happen at backend, then return that to client. What if client wants to display something else though? Should we be stuffing error codes into phrase_params? A little confused.

beyondcreed:
Hm, ok I guess that would show up in return structure. For type, maybe you misunderstood what I was trying to say about error_code. Its not really a group indicator, its more of a unique/specific code representing that exact error state or value. e.g. e_register_duplicate_email. But I guess type could be used for groupings of phrases with some kind of default group? So from what I see, we would stuff error codes into phrase params, then read on client from returned structure?

jhthorsen:
This is replaced by https://github.com/mojolicious/json-validator/pull/133

PR #132, opened 2018-12-12
Add value to enum validation error message

CandyAngel:
### Summary
Improve the error message for enum validation failures

### Motivation
The error message did not contain the value that had failed validation, making it more difficult than needed to either find the item with the incorrect data or add it to the enum if it was actually valid.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -683,7 +683,8 @@ sub _validate_type_enum {
   }
 
   local $" = ', ';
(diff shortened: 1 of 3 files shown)


jhthorsen:
I've considering this as well, but if this change is made, then I guess all of the other error messages need to include the input value as well.

I'm voting against this change for now.

jhthorsen:
Closing this, since no other comments/updated has been made.

Issue #129, opened 2018-12-11
adding error code to JSON::Validator::Error


beyondcreed:
just curious if it would be possible to add error_code to JSON::Validator::Error

e.g.

```
use JSON::Validator::Error;
my $err = JSON::Validator::Error->new($path, $message, $error_code);
```
use case for this would be to localize on front-end the message based on specific error code returned by API, could likely use message to stuff the code, but thought it would be cleaner to keep it separate.

beyondcreed:
sure, I'll head over there

jhthorsen:
How would the error code help you? Can you give me some examples of what they might be and how you would use them?

beyondcreed:
I commented on #130 with sample use case. you can check there.

jhthorsen:
Could you also have a look at https://github.com/mojolicious/json-validator/pull/133 to see what you think about that PR?

Issue #86, opened 2017-11-28
Internalization of validation messages


vividsnow:

jhthorsen:
I'm going to close this issue unless someone contributes or creates a plan for how this can be accomplished. Also, I want to know why this is important and which problem/usecase it fixes.

vividsnow:
Hello

I propose to use Locale::Maketext in statements like these:
https://github.com/jhthorsen/json-validator/blob/master/lib/JSON/Validator.pm#L174
https://github.com/jhthorsen/json-validator/blob/master/lib/JSON/Validator.pm#L497

This feature will give ability to display localized validation errors in UI

jhthorsen:
You should rather translate the errors in your UI. Also, displaying these errors directly to the end user will be confusing, unless they are extremely technical.

PR #133, opened 2018-12-15
Add JSON::Validator::Error->details()

jhthorsen:
### Summary
This change involves adding `details()` to the error object, so that you can inspect and a data structure to see what went wrong, instead of looking at `message()`

### Motivation
The motivation behind this is to give the users of JSON::Validator a chance to investigate the cause of an error programatically.

### References
* https://github.com/mojolicious/json-validator/pull/86
* https://github.com/mojolicious/json-validator/issues/129
* https://github.com/mojolicious/json-validator/pull/130

### DIFF
--- a/lib/JSON/Validator/Error.pm
+++ b/lib/JSON/Validator/Error.pm
@@ -3,16 +3,92 @@ use Mojo::Base -base;
 
 use overload q("") => \&to_string, bool => sub {1}, fallback => 1;
 
+our %MESSAGES = (
+  allOf => {type => '/allOf Expected %3 - got %4.'},
+  anyOf => {type => '/anyOf Expected %3 - got %4.'},
+  array => {
+    additionalItems => 'Invalid number of items: %3/%4.',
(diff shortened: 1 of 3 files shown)


lib/JSON/Validator/Error.pm
```diff
@@ -34,11 +110,44 @@ L<JSON::Validator>.
 
 =head1 ATTRIBUTES
 
+=head2 details
```
karenetheridge: I find this style of documentation to be really confusing. It's implying that the user should be assigning to details -- when actually, the user will only calling `details` as a getter, and will be returned an arrayref as a result. Indeed even the internals never use `details` as a setter -- this attribute is populated directly by the constructor.

jhthorsen: But it is perfectly fine to assign to it, if you want to make your own custom error, but in most cases you only want to read from it. I do welcome a better description that mention that though, but I fail to make up a short enough version.

PR #135, opened 2018-12-17
Cleaner blog example

jhthorsen:
### Summary
This PR tries to clean up the blog example.

### Motivation
@kraih gave me some feedback in #mojo.

### DIFF
--- a/examples/blog/lib/Blog.pm
+++ b/examples/blog/lib/Blog.pm
@@ -2,7 +2,8 @@ package Blog;
 use Mojo::Base 'Mojolicious';
 
 use Blog::Model::Posts;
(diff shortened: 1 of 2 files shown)


kraih:
While the example application gets a bit better with the changes, the file backend is still a big distracting hack with the serialization. Why do you want to teach that?

jhthorsen:
The example illustrates that even the files are all strings, then joi() makes sure that the output structure also contains true booleans and numbers.

jhthorsen:
This might sound a bit weird, but I'm going to merge this PR and then move the whole example into a separate branch. The reason for this is that I think the changes are an improvement, but the example as a whole has gotten some critique.

PR #146, opened 2019-02-07
Add support for OpenAPI3 nullable property

augensalat:
### Summary
This is an attempt to implement support for OpenAPI3 nullable property.
Beware! It is a "works-for-me" patch.

### Motivation
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-20
is currently broken in `JSON::Validator`.

### References
https://github.com/jhthorsen/mojolicious-plugin-openapi/issues/106

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -550,6 +550,8 @@ sub _validate {
   $data = $$to_json if $to_json;
   $type = $schema->{type} || _guess_schema_type($schema, $data);
 
+  return if !defined $data and $schema->{nullable};
+
   # Test base schema before allOf, anyOf or oneOf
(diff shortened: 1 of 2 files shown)


jhthorsen:
I don't want this change to be applied to JSON::Validator. Can you apply it to https://github.com/jhthorsen/mojolicious-plugin-openapi/blob/master/lib/JSON/Validator/OpenAPI/Mojolicious.pm instead?

jhthorsen:
This PR could've closed https://github.com/jhthorsen/mojolicious-plugin-openapi/issues/106, but I still don't think it's a good idea to add this to JV, when it's OpenAPI specific.

Issue #68, opened 2017-07-22
Email format validation is SCARY


rolosworld:
Test code:
https://pastebin.com/Ya07Qv9e

Results:
```
' or email='a@a.a: JSON::Validator | -
a@' or email='a.a: JSON::Validator | -
a@a.' or email='a: JSON::Validator | -
' or email='a@' or email='a.' or email='a: JSON::Validator | -
 @a.a: - | -
a..a@a.a: JSON::Validator | -
a.@a.a: - | -
.a@a.a: JSON::Validator | -
a@.a.a: - | -
a@a..a: JSON::Validator | -
a@a.a.: JSON::Validator | -
.a@a.a.: JSON::Validator | -
a@a.a: JSON::Validator | Email::Valid
email@example.com: JSON::Validator | Email::Valid
firstname.lastname@example.com: JSON::Validator | Email::Valid
email@subdomain.example.com: JSON::Validator | Email::Valid
firstname+lastname@example.com: JSON::Validator | Email::Valid
email@123.123.123.123: JSON::Validator | Email::Valid
email@[123.123.123.123]: JSON::Validator | Email::Valid
"email"@example.com: JSON::Validator | Email::Valid
1234567890@example.com: JSON::Validator | Email::Valid
email@example-one.com: JSON::Validator | Email::Valid
_______@example.com: JSON::Validator | Email::Valid
email@example.name: JSON::Validator | Email::Valid
email@example.museum: JSON::Validator | Email::Valid
email@example.co.jp: JSON::Validator | Email::Valid
firstname-lastname@example.com: JSON::Validator | Email::Valid
much."more\ unusual"@example.com: JSON::Validator | Email::Valid
very.unusual."@".unusual.com@example.com: JSON::Validator | Email::Valid
very."(),:;<>[]".VERY."very@\\\ \"very".unusual@strange.example.com: JSON::Validator | Email::Valid
plainaddress: - | -
#@%^%#$@#$@#.com: JSON::Validator | -
@example.com: - | -
Joe Smith <email@example.com>: JSON::Validator | Email::Valid
email.example.com: - | -
email@example@example.com: JSON::Validator | -
.email@example.com: JSON::Validator | -
email.@example.com: - | -
email..email@example.com: JSON::Validator | -
あいうえお@example.com: - | -
email@example.com (Joe Smith): JSON::Validator | Email::Valid
email@example: JSON::Validator | -
email@-example.com: JSON::Validator | -
[...]

PR #148, opened 2019-02-15
Created method JSON::Validator->set_format_validator

rolosworld:
The method JSON::Validator->set_format_validator lets you set custom format validators or replace a default one.

### Summary
To achieve my objective I moved $_default_formats hash outside the method JSON::Validator->_build_formats to prevent new hash from being created everytime the method gets called.

Then modified JSON::Validator->_build_formats to make a copy of the hash on the object that called the method. The reason I made this change is so each object created can have their own format validators.

FInally, I created the method JSON::Validator->set_format_validator which simply sets the validator subroutine for the given format on the current object.

### Motivation
While using the module in a personal project I had some issues with how the email and uri formats were validated. I managed to create a workaround by overriding the JSON::Validator->_build_format method with modules that specialize validating those formats (Email::Valid and Data::Validate::URI) . I figured it could be useful to change the format validators, so here we are.

### References
https://github.com/mojolicious/json-validator/issues/68

[test.pl.txt](https://github.com/mojolicious/json-validator/files/2871130/test.pl.txt)

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -193,8 +193,7 @@ sub validate_json {
   __PACKAGE__->singleton->schema($_[1])->validate($_[0]);
 }
 
-sub _build_formats {
-  return {
+my $_default_formats = {
     'date'          => JSON::Validator::Formats->can('check_date'),
     'date-time'     => JSON::Validator::Formats->can('check_date_time'),
     'email'         => JSON::Validator::Formats->can('check_email'),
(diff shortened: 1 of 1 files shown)


jhthorsen:
Even though there’s no custom method for adding/changing formats, but you can already just manipulate the “formats” attribute, which holds all of them in a hash or create new objects with the attribute set in the constructor.

A shared variable won’t work in the global scope, since changing the “formats” attribute would the change every object.

You can discuss what you’re trying to achieve in #perl-openapi channel on the freenode IRC server or follow up with a comment here, but I won’t take the PR as it stands now.

When that is said: I’m more interested in two new issues that explains which email and uri formats is not supported, since that sounds like a bug that needs fixing. But make sure you’re not mixing up “iri” and “uri” first :)

rolosworld:
> When that is said: I’m more interested in two new issues that explains which email and uri formats is not supported, since that sounds like a bug that needs fixing. But make sure you’re not mixing up “iri” and “uri” first :)

Sorry, right now I don't remember exactly why I modified the "uri" format validator, I made the change a while back and didn't make a note of what issue I had.

About the reason I modified the "email" format validator, when using the default validator it says these emails are valid:

- ' or email='a@a.a
- a@' or email='a.a
- a@a.' or email='a
- ' or email='a@' or email='a.' or email='a
- a..a@a.a
- .a@a.a
- a@a..a
- a@a.a.
- .a@a.a.
- #@%^%#$@#$@#.com
- email@example@example.com
- .email@example.com
- email..email@example.com
- email@example
- email@-example.com
- email@example..com
- Abc..123@example.com
- just"not"right@example.com
- this\ is\"really\"not\\allowed@example.com

You can review + run the test.pl script I provided and confirm it. I don't know if those emails are valid rfc5322 emails, but these are changes I needed and I wasn't restricted by any specifications.

PR #151, opened 2019-03-18
pod and cache test fixes

karenetheridge:
- minor pod markup fixes
- fix a test that seems to have been overlooked because it doesn't normally run: an old environment variable was being used, and the spec URL no longer worked.

replaces parts of #150.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -1294,7 +1294,7 @@ Used to create a new schema, where the C<$ref> are resolved. C<%args> can have:
 
 =over 2
 
-=item * C<{replace => 1}>
+=item * C<< {replace => 1} >>
 
 Used if you want to replace the C<$ref> inline in the schema. This currently
 does not work if you have circular references. The default is to move all the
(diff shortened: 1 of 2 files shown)


lib/JSON/Validator.pm
```diff
@@ -1294,7 +1294,7 @@ Used to create a new schema, where the C<$ref> are resolved. C<%args> can have:
 
 =over 2
 
-=item * C<{replace => 1}>
+=item * C<< { replace => 1 } >>
```
jhthorsen: I don’t fancy the extra white space. Please remove it so it looks like this: `<<{replace => 1}>>`.

karenetheridge: done.

lib/JSON/Validator.pm
```diff
@@ -1307,7 +1307,7 @@ on how a C<$ref> looks before and after:
   {"$ref":"http://example.com#/foo/bar"}
      => {"$ref":"#/definitions/_http___example_com-_foo_bar"}
 
-=item * C<{schema => {...}}>
+=item * C< {schema => {...}} >>
```
jhthorsen: There seems to be a typo here? (Missing <)

Also, is the white space requires about {}?

karenetheridge: Typo fixed! sorry about that.

Yes, the tags are `C<< ` and ` >>C`. The whitespace isn't visible. https://perldoc.perl.org/perlpod.html#Formatting-Codes

jhthorsen:
Thanks!

PR #152, opened 2019-03-18
Extract $schema from document

karenetheridge:
Also infer the schema from the "$schema" key in the document, when available.  Previously, when using a schema document that was downloaded from an external source, only the `$args` variable was used to determine the schema's version; we did not also look inside the document for a `$schema` key.  Now we do both, deferring the second check until after the document has been downloaded and parsing begins.

replaces parts of #150.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,8 @@
 Revision history for perl distribution JSON-Validator
 
+ - load_and_validate_schema() now also looks for the schema version in the "$schema"
+   key of the provided document
+
 3.11 2019-05-07T21:53:16+0700
  - Bundle https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.json #157
 
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -157,9 +157,10 @@ sub joi {
 
 sub load_and_validate_schema {
   my ($self, $spec, $args) = @_;
-  my $schema = $args->{schema} || SPECIFICATION_URL;
-  $self->version($1) if !$self->{version} and $schema =~ /draft-0+(\w+)/;
+  my $schema = $args->{schema} || (ref $spec eq 'HASH' && $spec->{'$schema'});
+  $self->version($1) if $schema and !$self->{version} and $schema =~ /draft-0+(\w+)/;
   $spec = $self->_resolve($spec);
+  $schema ||= $spec->{'$schema'} || SPECIFICATION_URL;
   my @errors = $self->new(%$self)->schema($schema)->validate($spec);
   confess join "\n", "Invalid JSON specification $spec:", map {"- $_"} @errors
     if @errors;
@@ -443,14 +444,13 @@ sub _report_schema {
 # resolve all the $ref's that we find inside JSON Schema specification.
 sub _resolve {
   my ($self, $schema) = @_;
-  my $id_key = $self->_id_key;
   my ($id, $resolved, @refs);
 
   local $self->{level} = $self->{level} || 0;
   delete $_[0]->{schemas}{''} unless $self->{level};
 
   if (ref $schema eq 'HASH') {
-    $id = $schema->{$id_key} // '';
+    $id = $schema->{$self->_id_key} // '';
     return $resolved if $resolved = $self->{schemas}{$id};
   }
   elsif ($resolved = $self->{schemas}{$schema // ''}) {
@@ -458,8 +458,12 @@ sub _resolve {
   }
   else {
     ($schema, $id) = $self->_load_schema($schema);
+    $self->version($1) if !$self->{version} and $schema->{'$schema'} and $schema->{'$schema'} =~ /draft-0+(\w+)/;
+    my $id_key = $self->_id_key;
     $id = $schema->{$id_key} if $schema->{$id_key};
   }
+
+  my $id_key = $self->_id_key;
 
   unless ($self->{level}) {
     my $rid = $schema->{$id_key} // $id;
--- a/t/id-keyword-draft6.t
+++ b/t/id-keyword-draft6.t
@@ -1,0 +1,43 @@
+use Mojo::Base -strict;
+use Test::Mojo;
+use Test::More;
+use JSON::Validator;
+
+use Mojolicious::Lite;
+get '/person'           => 'person';
+get '/invalid-relative' => 'invalid-relative';
+
+my $t  = Test::Mojo->new;
+my $jv = JSON::Validator->new(ua => $t->ua);
+
+$t->get_ok('/person.json')->status_is(200);
+my $base_url = $t->tx->req->url->to_abs->path('/');
+eval {
+  $jv->load_and_validate_schema("${base_url}person.json",
+    {schema => 'http://json-schema.org/draft-06/schema'});
+};
+ok !$@, "${base_url}schema.json" or diag $@;
+
+is $jv->version, 6,    'detected version from $arg, as draft-06';
+is $jv->_id_key, 'id', 'detected id_key from $arg, as draft-06';
+
+eval { $jv->load_and_validate_schema("${base_url}invalid-relative.json") };
+ok !$@, 'Root id can be relative' or diag $@;
+
+done_testing;
+
+__DATA__
+@@ invalid-relative.json.ep
+{"$id": "whatever"}
+@@ person.json.ep
+{
+  "$id": "http://example.com/person.json",
+  "definitions": {
+    "Person": {
+      "type": "object",
+      "properties": {
+        "firstName": { "type": "string" }
+      }
+    }
+  }
+}
--- a/t/version-detection.t
+++ b/t/version-detection.t
@@ -1,0 +1,51 @@
+use Mojo::Base -strict;
+use Test::Mojo;
+use Test::More;
+use JSON::Validator;
+
+use Mojolicious::Lite;
+get '/person' => 'person';
+
+my $t = Test::Mojo->new;
+$t->get_ok('/person.json')->status_is(200);
+my $base_url = $t->tx->req->url->to_abs->path('/');
+my $document = $t->tx->res->json;
+
+{
+  my $jv = JSON::Validator->new;
+
+  eval { $jv->load_and_validate_schema($document); };
+  ok !$@, 'local schema.json' or diag $@;
+
+  is $jv->version, 7,     'detected version from local document, as draft-07';
+  is $jv->_id_key, '$id', 'detected id_key from local document, as draft-07';
+}
+
+{
+  my $jv = JSON::Validator->new(ua => $t->ua);
+
+  eval { $jv->load_and_validate_schema("${base_url}person.json"); };
+  ok !$@, "${base_url}schema.json" or diag $@;
+
+  is $jv->version, 7,     'detected version from document via url, as draft-07';
+  is $jv->_id_key, '$id', 'detected id_key from document via url, as draft-07';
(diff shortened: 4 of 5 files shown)


karenetheridge:
rebased to master after the last PR got merged.

Changes
```diff
@@ -1,5 +1,8 @@
 Revision history for perl distribution JSON-Validator
 
+ - load_and_validate_schema() now also looks for the schema version in the "$schema"
+   key of the provided document
```
jhthorsen: I'll take care of "Changes". You can remove this change.

lib/JSON/Validator.pm
```diff
@@ -155,9 +155,10 @@ sub joi {
 
 sub load_and_validate_schema {
   my ($self, $spec, $args) = @_;
-  my $schema = $args->{schema} || SPECIFICATION_URL;
-  $self->version($1) if !$self->{version} and $schema =~ /draft-0+(\w+)/;
+  my $schema = $args->{schema} || (ref $spec eq 'HASH' && $spec->{'$schema'});
+  $self->version($1) if $schema and !$self->{version} and $schema =~ /draft-0+(\w+)/;
   $spec = $self->_resolve($spec);
+  $schema ||= $spec->{'$schema'} || SPECIFICATION_URL;
```
jhthorsen: I don't understand this change: Haven't you already picked out `$schema` from the spec above? I don't see how `_resolve()` can alter this part of `$spec`. Also, what about fetching `version()` from `SPECIFICATION_URL`?

lib/JSON/Validator.pm
```diff
@@ -407,24 +408,27 @@ sub _report_schema {
 # resolve all the $ref's that we find inside JSON Schema specification.
 sub _resolve {
```
jhthorsen: I'm not a huge fan of this change, since `_resolve()` is called from multiple places in the code, which might result in unexpected results, dependent on when `version()` is set the first time.

t/id-keyword-draft6.t
```diff
@@ -0,0 +1,45 @@
...
+use JSON::Validator;
+
+my ($base_url, $jv, $t, @e);
+
+use Mojolicious::Lite;
+get '/person'           => 'person';
+get '/invalid-relative' => 'invalid-relative';
+
+$t = Test::Mojo->new;
+$jv = JSON::Validator->new(ua => $t->ua);
```
jhthorsen: I don't think the code (any of the files in this PR) is run through perltidy. Please do so, using the perltidyrc found in this project.

t/id-keyword-draft6.t
```diff
@@ -0,0 +1,45 @@
...
+  $jv->load_and_validate_schema("${base_url}person.json",
+    {schema => 'http://json-schema.org/draft-06/schema'});
+};
+ok !$@, "${base_url}schema.json" or diag $@;
+
+is $jv->version, 6,     'detected version from $arg, as draft-06';
+is $jv->_id_key, 'id',  'detected id_key from $arg, as draft-06';
+
+eval { $jv->load_and_validate_schema("${base_url}invalid-relative.json") };
+is $@, '', 'Root id can be relative' or diag $@;
```
jhthorsen: I like the `ok !$@, ...` test above better.

t/id-keyword-draft6.t
```diff
@@ -0,0 +1,45 @@
...
+use Mojolicious::Lite;
+get '/person'           => 'person';
+get '/invalid-relative' => 'invalid-relative';
+
+$t = Test::Mojo->new;
+$jv = JSON::Validator->new(ua => $t->ua);
+
+eval {
+  $t->get_ok('/person.json')->status_is(200);
+  $base_url = $t->tx->req->url->to_abs->path('/');
```
jhthorsen: There's a little bit too much things inside this eval. The two lines above should never fail.

t/id-keyword-draft6.t
```diff
@@ -0,0 +1,45 @@
+use Mojo::Base -strict;
+use Test::Mojo;
+use Test::More;
+use JSON::Validator;
+
+my ($base_url, $jv, $t, @e);
```
jhthorsen: I think it's better if you declare this variables where you first assign to them. I also don't think all of them are in use?

t/version-detection.t
```diff
@@ -0,0 +1,56 @@
+use Mojo::Base -strict;
+use Test::Mojo;
+use Test::More;
+use JSON::Validator;
+
+use Mojolicious::Lite;
+get '/person'           => 'person';
+
+my $t = Test::Mojo->new;
+$t->get_ok('/person.json')->status_is(200);
```
jhthorsen: You don't really have to test this. It's done implicit by `load_and_validate_schema()`.

t/version-detection.t
```diff
@@ -0,0 +1,56 @@
...
+
+  is $jv->version, 7,     'detected version from local document, as draft-07';
+  is $jv->_id_key, '$id', 'detected id_key from local document, as draft-07';
+}
+
+{
+  my $jv = JSON::Validator->new(ua => $t->ua);
+
+  eval {
+    $jv->load_and_validate_schema("${base_url}person.json");
```
jhthorsen: You shouldn't need `$base_url`, since you pass `$t->ua` into `new()`.

t/id-keyword-draft6.t
karenetheridge: I tried to keep it as close to the existing t/id-keyword-draft7.t as possible :p

t/id-keyword-draft6.t
karenetheridge: I copied this file from t/id-keyword-draft7.t thinking that this was the style you preferred. Shall I fix that file as well?

t/version-detection.t
karenetheridge: Again all of this was copied directly from what you did in t/id-keyword-draft7.t. I am not a big fan of the style; it is very different from what I would have done in my own code. I copied it thinking it was the style you preferred.

lib/JSON/Validator.pm
karenetheridge: It gets tricky. The changes I made are the easiest way I found to handle all the different usecases.  This is exactly why I moved the `$self->_id_key` call at the top of `_resolve` to lower down, to inside the `ref $schema eq 'HASH'` section where we know it's safe (for hashref schemas, we have already fetched the `'$schema'` key and therefore it is safe to call `$self->$self->version` and `$self->id_key`. For the URL/filename case we need to delay calculating the version until after we've fetched the file and can look inside its contents.

lib/JSON/Validator.pm
karenetheridge: We can extract `'$schema'` from the spec here (line 158) only if the spec was a hashref. If it was provided as a URL, we don't get to see inside it until after we've run _resolve, so when that's returned, then we can look at `$spec->{'$schema'}` and set it if it wasn't already.  _resolve is where the spec is downloaded (or loaded from a local file).

t/version-detection.t
karenetheridge: It is apparently needed; tests fail when `$base_url` is removed entirely. That seems to be the reason why the `$t->get_ok` line exists as well.

karenetheridge:
I've pushed some amendments as a new commit, that I'm happy to rebase back into the original commits later (if I squashed them now, all the review comments above are much harder to see in github's UI).

t/id-keyword-draft6.t
jhthorsen: That would be awesome. My style when declaring variables used to be quite consistent, but the last year, I've changed my style, so it might be a mess multiple places in the code 😞

karenetheridge: ok, I pushed a commit to tidy up t/id-keyword-draft7.t.

lib/JSON/Validator.pm
jhthorsen: It might be very pretty, but I think we should do this:

```
diff --git a/lib/JSON/Validator.pm b/lib/JSON/Validator.pm
index e9a7927..cacf307 100644
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -161,7 +161,7 @@ sub load_and_validate_schema {
   my ($self, $spec, $args) = @_;
   my $schema = $args->{schema} || SPECIFICATION_URL;
   $self->version($1) if !$self->{version} and $schema =~ /draft-0+(\w+)/;
-  $spec = $self->_resolve($spec);
+  $spec = $self->_resolve($spec, 1);
   my @errors = $self->new(%$self)->schema($schema)->validate($spec);
   confess join "\n", "Invalid JSON specification $spec:", map {"- $_"} @errors
     if @errors;
@@ -178,7 +178,7 @@ sub new {
 sub schema {
   my $self = shift;
   return $self->{schema} unless @_;
-  $self->{schema} = Mojo::JSON::Pointer->new($self->_resolve(shift));
+  $self->{schema} = Mojo::JSON::Pointer->new($self->_resolve(shift, 1));
   return $self;
 }
 
@@ -442,7 +442,7 @@ sub _report_schema {
 # _resolve() method is used to convert all "id" into absolute URLs and
 # resolve all the $ref's that we find inside JSON Schema specification.
 sub _resolve {
-  my ($self, $schema) = @_;
+  my ($self, $schema, $detect_version) = @_;
   my $id_key = $self->_id_key;
   my ($id, $resolved, @refs);
```

By doing that, we can use the `$detect_version` flag to tell _resolve() that it is in fact allowed to set `version()`.

What do you think?

Note that the diff above is not based on your changes. I apologise for that.

jhthorsen:
@karenetheridge: Got any feedback on https://github.com/mojolicious/json-validator/pull/152/files#r272815748 ?

karenetheridge:
> @karenetheridge: Got any feedback on https://github.com/mojolicious/json-validator/pull/152/files#r272815748 ?

I missed that before! I will look at ASAP (today/tomorrow).

lib/JSON/Validator.pm
karenetheridge: In your diff, you've still got schema being extracted wrong -- it's important to not default it to SPECIFICATION_URL until after we've looked inside `$spec` to see if `$spec->{'$schema'}` is set.

So, I can see that being explicit could be better, but I'm not sure if it will help anything here -- we can only set version if the schema URL has been determined, and to do that we need to look inside the data. That's why I did that right at the beginning, at line 158, before even calling _resolve.

karenetheridge: The part I don't like about my patch is that it does this code twice (and in two different functions even):

    $self->version($1) if !$self->{version} and $schema->{'$schema'} and $schema->{'$schema'} =~ /draft-0+(\w+)/;

..so it would be good to do that just once... the problem is there are so many different ways the schema could be defined, there are a lot of cases to cover.  I suspect some other code needs to move around as well to make everything more consistent.

jhthorsen:
You might want to have a quick look at https://github.com/mojolicious/json-validator/compare/jhthorsen/strict. I want to make JSON::Validator strict by default now, but doing so without the user noticing - hopefully. I think I will scrap that branch, but wanted to show it for inspiration (?).

jhthorsen:
The more I look at the code, the more I think I want to get rid of `load_and_validate_schema()`. I started on another suggestion [here](https://github.com/mojolicious/json-validator/compare/jhthorsen/refactor) which has `new_from_schema()` which again will find the appropriate "Draft" subclass. I'll let it sit for a bit, but I will hopefully find time to work on this later this week.

jhthorsen:
I'm going to close this in favor of #189.

PR #156, opened 2019-05-06
Access to schema fragment from custom format codeblocks

jmfernandez:
### Summary
Custom formats codeblocks now receive as second parameter the schema fragment, in case additional attributes are needed to tweak the validation, per schema fragment.

### Motivation
In some of our projects we want to tweak custom formats validation based on values from the schema. 

### References
(none)

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -746,7 +746,7 @@ sub _validate_format {
   my $code = $self->formats->{$schema->{format}};
   return do { warn "Format rule for '$schema->{format}' is missing"; return }
     unless $code;
-  return unless my $err = $code->($value);
+  return unless my $err = $code->($value,$schema);
   return E $path, $err;
 }
 
@@ -1291,7 +1291,10 @@ See L</Bundled specifications> for more details.
   my $jv = $jv->formats(\%hash);
 
 Holds a hash-ref, where the keys are supported JSON type "formats", and
-the values holds a code block which can validate a given format. A code
+the values holds a code block which can validate a value against a given
+format. The second parameter is the schema fragment where the format is
+defined, in case some custom format requires accessing additional schema
+attributes in order to tweak its validation. A code
 block should return C<undef> on success and an error string on error:
 
   sub { return defined $_[0] && $_[0] eq "42" ? undef : "Not the answer." };
--- a/t/Helper.pm
+++ b/t/Helper.pm
@@ -38,6 +38,22 @@ sub validate_ok {
     or Test::More::diag(encode_json(\@errors));
 }
 
+sub validate_custom_format_ok {
+  my ($data, $schema, $newformat, $codeblock, @expected) = @_;
+  my $description
+    ||= @expected ? "errors: @expected" : "valid: " . encode_json($data);
+  my $jv = jv()->schema($schema);
+  my $formats = $jv->formats;
+  $formats->{$newformat} = $codeblock;
+  
+  my @errors = $jv->formats($formats)->validate($data);
+  Test::More::is_deeply(
+    [map { $_->TO_JSON } sort { $a->path cmp $b->path } @errors],
+    [map { $_->TO_JSON } sort { $a->path cmp $b->path } @expected],
+    $description)
+    or Test::More::diag(encode_json(\@errors));
+}
+
 sub import {
   my $class  = shift;
   my $caller = caller;
@@ -52,6 +68,7 @@ sub import {
   monkey_patch $caller => jv           => \&jv;
   monkey_patch $caller => true         => \&Mojo::JSON::true;
   monkey_patch $caller => validate_ok  => \&validate_ok;
+  monkey_patch $caller => validate_custom_format_ok  => \&validate_custom_format_ok;
 }
 
 1;
--- a/t/custom-formats.t
+++ b/t/custom-formats.t
@@ -3,43 +3,47 @@ use t::Helper;
 use Mojo::Util 'decode';
 use Test::More;
 
-my $schema = {type => 'object', properties => {v => {type => 'string'}}};
+my $palindrome_schema = {
+  type       => 'object',
+  properties => {v => {type => 'string', format => 'palindrome'}}
+};
 
 {
-  local $schema->{properties}{v}{format} = 'palindrome';
-  
   my $check_palindrome = sub {
     return $_[0] eq reverse($_[0]) ? undef : 'Not a palindrome.';
   };
-  
-  validate_custom_format_ok {v => 'Able was I ere I saw elbA'}, $schema, 'palindrome', $check_palindrome;
-  
-  validate_custom_format_ok {v => 'Amoeba'}, $schema,
-    'palindrome', $check_palindrome, E('/v', 'Not a palindrome.');
+
+  validate_custom_format_ok {v => 'Able was I ere I saw elbA'},
+    $palindrome_schema, 'palindrome', $check_palindrome;
+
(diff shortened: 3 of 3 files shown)


jmfernandez:
I have just added a new test file for custom formats, including the extension proposed in this pull request.

jhthorsen:
I'm still sceptical to the change, even though I can see the usefulness in the tests. Do you have an example of library written in a different language that does this?

Also: You still haven't run the code through perltidy.

jhthorsen:
I kind of want to break the custom validation again and do:

```
$self->$code($value, $path, $schema);
```

...but I guess it's not worth it.

I'll merge this after it's tidied up and if no-one else objects. (Let me know if you need help running perltidy)

jmfernandez:
Yes, please. I haven't used perltidy before, so I don't know the steps you usually follow

karenetheridge:
@jhthorsen I think it would be helpful for users if you created a 'CONTRIBUTING' file in your repository -- this will be shown to users when they create a pull request. In it you can put your guidelines for creating good patches, such as how to run perltidy using the config file you have created -- as this requirement is not presently made known to people before they submit and it catches them by surprise.

t/custom-formats.t
```diff
@@ -0,0 +1,45 @@
+use lib '.';
+use t::Helper;
+use Mojo::Util 'decode';
+use Test::More;
+
+my $schema = {type => 'object', properties => {v => {type => 'string'}}};
+
+{
+  local $schema->{properties}{v}{format} = 'palindrome';
```
karenetheridge: Instead of using 'local' to patch a single schema, I would just create a new document here, which would make it easier to see what it looks like as a whole.

lib/JSON/Validator.pm
```diff
@@ -1291,7 +1291,10 @@ See L</Bundled specifications> for more details.
   my $jv = $jv->formats(\%hash);
 
 Holds a hash-ref, where the keys are supported JSON type "formats", and
-the values holds a code block which can validate a given format. A code
+the values holds a code block which can validate a value against a given
```
karenetheridge: side observation, to @jhthorsen -- it doesn't look like there are any tests in the distribution for the user adding his own formats to the validator -- e.g. `$jv->formats->{my_custom_format} = sub { ... }`. I didn't realize this was supported until I saw this PR!

t/custom-formats.t
jhthorsen: I think this is probably written this way, since I've done that in other tests. Not saying it's the best way, but it saves a lot of keystrokes.

jhthorsen:
@jmfernandez: I use `githook-perltidy install` which create git hooks that tidy the code for me when I commit. So I would suggest installing that (it will only affect the current repo you run it in) and then add a random space character inside the files and then do `git commit -a -m"tidy the code"`.

https://metacpan.org/pod/distribution/App-githook-perltidy/bin/githook-perltidy#githook-perltidy-install-[-force,-f]

lib/JSON/Validator.pm
jhthorsen: @karenetheridge: If that didn't work, then none of the format tests would work. I'm not encouraging adding custom formats though, since I'm not sure if that will actually result in a portable schema.

t/custom-formats.t
jmfernandez: Sorry, past month was very stressing. I have just followed your recommendation @karenetheridge  in https://github.com/fairtracks/json-validator/commit/94cba172282bb1610c250413d01c5a651dd3a12c

jmfernandez:
Sorry, past month was very stressing. @jhthorsen I have just had the chance to install githook-perltidy and its hook. After fixing an issue with a locally installed stale perltidy (two years ago version), I was able to commit again the changes (see https://github.com/fairtracks/json-validator/commit/94cba172282bb1610c250413d01c5a651dd3a12c)

lib/JSON/Validator.pm
```diff
@@ -745,7 +745,7 @@ sub _validate_format {
   my $code = $self->formats->{$schema->{format}};
   return do { warn "Format rule for '$schema->{format}' is missing"; return }
     unless $code;
-  return unless my $err = $code->($value);
+  return unless my $err = $code->($value,$schema);
```
jhthorsen: I think I want this to change after all to `($value, $schema, $path)` instead of just `($value, $schema)`. This makes the callback receive all the local information that `_validate_format()` gets. The order is different though, but I think the most common case would be to need $schema and not $path.

What do you think?

Also, this line is not run through perltidy. I think you can now run the command below after 316bf1832dacdb686d0fe81ba20c87d390c950fc.

```
perltidy -b lib/JSON/Validator.pm
```

lib/JSON/Validator.pm
```diff
@@ -1290,7 +1290,10 @@ See L</Bundled specifications> for more details.
   my $jv = $jv->formats(\%hash);
 
 Holds a hash-ref, where the keys are supported JSON type "formats", and
-the values holds a code block which can validate a given format. A code
+the values holds a code block which can validate a value against a given
+format. The second parameter is the schema fragment where the format is
+defined, in case some custom format requires accessing additional schema
+attributes in order to tweak its validation. A code
```
jhthorsen: I think this documentation is very confusing now. I suggest trying to illustrate how it works with a better code example. Something like this:

```
=head2 formats
  
  my $hash_ref = $jv->formats;
  my $jv       = $jv->formats(\%hash);
  
Holds a hash-ref, where the keys are supported JSON type "formats", and the 
values holds a code block which can validate a given format. The code block
must return C<undef> on success and a string on error:
  
  $jv->formats->{custom_format} = sub {
    my $value                   = shift; # User input
    my $schema_fragment         = shift; # Example: {type => "string", format => "custom_format"}
    my $path_to_schema_fragment = shift; # Example "/some/location/in/schema"
  
    return defined $value && $value eq "42" ? undef : "Not 42.";
  };  
  
See L<JSON::Validator::Formats> for a list of supported formats.
```

t/Helper.pm
```diff
@@ -38,6 +38,22 @@ sub validate_ok {
     or Test::More::diag(encode_json(\@errors));
 }
 
+sub validate_custom_format_ok {
```
jhthorsen: I don't really want this function here, since it's not a generic function that will be used in more test files.

t/custom-formats.t
```diff
@@ -0,0 +1,49 @@
...
+  my $check_palindrome = sub {
+    return $_[0] eq reverse($_[0]) ? undef : 'Not a palindrome.';
+  };
+
+  validate_custom_format_ok {v => 'Able was I ere I saw elbA'},
+    $palindrome_schema, 'palindrome', $check_palindrome;
+
+  validate_custom_format_ok {v => 'Amoeba'}, $palindrome_schema, 'palindrome',
+    $check_palindrome, E('/v', 'Not a palindrome.');
+}
```
jhthorsen: You can write this test in this way instead:

```
jv->formats->{palindrome} = sub {
  return $_[0] eq reverse($_[0]) ? undef : 'Not a palindrome.';
};

my $palindrome_schema = {
  type       => 'object',
  properties => {v => {type => 'string', format => 'palindrome'}}
};

validate_ok {v => 'Able was I ere I saw elbA'}, $palindrome_schema;
validate_ok {v => 'Amoeba'}, $palindrome_schema, $check_palindrome,
  E('/v', 'Not a palindrome.');
```

You can skip the block around the test, since none of the tests are overlapping.

jhthorsen:
I'm going to close this PR, even though I think it's a good idea. I'm closing it because it's in conflict with master and it seems to have gone stale.

PR #159, opened 2019-05-15
Add more tests for bundle()

karenetheridge:
These are all the tests I was using for my implementation of bundle(), including a few interesting cases of name duplication, recursive references, etc.  All of my test cases are now implemented by `bundle` as of version 3.11, so hooray :)

### DIFF
--- a/t/more-bundle.t
+++ b/t/more-bundle.t
@@ -1,0 +1,236 @@
+use Mojo::Base -strict;
+
+use Test::More;
+use Test::Deep;
+use JSON::Validator;
+
+# these are triples:
+# - schema name to extract from schema file(s) with bundle()
+# - expected result
+# - test name.
+my @tests = (
+  [
+    'i_have_nested_refs',
+    {
+      definitions => {
+        ref1 => {type => 'array', items => {'$ref' => '#/definitions/ref2'}},
+        ref2 => {type => 'string', minLength => 1},
+      },
+
+      # begin i_have_nested_refs definition
+      type       => 'object',
+      properties => {
+        my_key1 => {'$ref' => '#/definitions/ref1'},
+        my_key2 => {'$ref' => '#/definitions/ref1'},
+      },
+    },
+    'find and resolve nested $refs; main schema is at the top level',
+  ],
+
+  [
+    'i_have_a_recursive_ref',
+    {
+      definitions => {
+        i_have_a_recursive_ref => {
+          type       => 'object',
+          properties => {
+            name     => {type => 'string'},
+            children => {
+              type    => 'array',
+              items   => {'$ref' => '#/definitions/i_have_a_recursive_ref'},
+              default => [],
+            },
+          },
+        },
+      },
+
+      # begin i_have_a_recursive_ref definition
+      # it is duplicated with the above, but there is no other way,
+      # because $ref cannot be combined with other sibling keys
+      type       => 'object',
+      properties => {
+        name     => {type => 'string'},
+        children => {
+          type    => 'array',
+          items   => {'$ref' => '#/definitions/i_have_a_recursive_ref'},
+          default => [],
+        },
+      },
+    },
+    'find and resolve recursive $refs',
+  ],
+
+  [
+    'i_have_a_ref_to_another_file',
+    {
+      definitions => {
+        my_name    => {type => 'string', minLength => 2},
+        my_address => {
+          type       => 'object',
+          properties => {
+            street => {type   => 'string'},
+            city   => {'$ref' => '#/definitions/my_name'},
+          },
+        },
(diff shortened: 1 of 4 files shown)


t/more-bundle.t
```diff
@@ -0,0 +1,264 @@
...
+
+use Test::More;
+use Test::Deep;
+use JSON::Validator;
+
+# these are triples:
+# - schema name to extract from schema file(s) with bundle()
+# - expected result
+# - test name.
+my @tests = (
```
jhthorsen: I'm pretty sure none of the files are run through perltidy.

t/more-bundle.t
```diff
@@ -0,0 +1,264 @@
...
+            type => 'object',
+            properties => {
+                gotcha => { '$ref' => '#/definitions/ref3' },
+            },
+        },
+        'find and resolve a reference that immediately leaps to another file',
+    ],
+);
+
+my $_validator = JSON::Validator->new;
```
jhthorsen: I find it very confusing following $jv, $_jv and $_validator. Can we name them something more descriptive?

t/more-bundle.t
```diff
@@ -0,0 +1,264 @@
+use strict;
+use warnings;
```
jhthorsen: Please replace these two use statements with `use Mojo::Base -strict;`

t/more-bundle.t
```diff
@@ -0,0 +1,264 @@
...
+    cmp_deeply(
+        $_jv->schema->data,
+        $expected_output,
+        'our generated schema does not lose any data when parsed again by a validator',
+    );
+}
+foreach @tests;
+
+done_testing;
+# vim: set ts=4 sts=4 sw=4 et :
```
jhthorsen: I don't want $EDITOR config in the source files.

t/more-bundle.t
```diff
@@ -0,0 +1,264 @@
...
+    ],
+);
+
+my $_validator = JSON::Validator->new;
+$_validator->schema('http://json-schema.org/draft-07/schema#');
+
+my $jv = JSON::Validator->new;
+$jv->load_and_validate_schema('t/spec/more-bundle.yaml', { schema => 'http://json-schema.org/draft-07/schema#' });
+
+subtest $_->[2] => sub {
```
jhthorsen: I'm not a huge fan of subtests, since it makes it harder for me to figure out what has gone wrong when a smoker fail - at least I think that was my problem many years ago. Can you please change this to:

```
for my $t (@tests) {
  ...
}
```

t/more-bundle.t
```diff
@@ -0,0 +1,264 @@
...
+
+    my $_jv = JSON::Validator->new;
+    $_jv->load_and_validate_schema($got, { schema => 'http://json-schema.org/draft-07/schema#' });
+    cmp_deeply(
+        $_jv->schema->data,
+        $expected_output,
+        'our generated schema does not lose any data when parsed again by a validator',
+    );
+}
+foreach @tests;
```
jhthorsen: Please use `for` instead of `foreach`.

jhthorsen:
@karenetheridge: What's the status of this PR? I'll probably accept it and fix up the comments myself, if this PR is idling for much longer. (That means that I will be credited via "git blame" on many lines after running through perltidy)

karenetheridge:
I plan to come back to this next week.

t/more-bundle.t
karenetheridge: correct.

t/more-bundle.t
karenetheridge: I'll rename the variables, to indicate what they're each used for.

t/more-bundle.t
karenetheridge: The intent of the subtests is to make it *easier* to see what went wrong -- the tests are all the same for each block of data, so the subtest is labeled for which block of data is being tested.  Please take a look at the output again to see how much clearer it is with each dataset in its own section.

karenetheridge:
I've force-pushed some changes.

t/more-bundle.t
jhthorsen: Let me rephrase: All the code need to be run through "perltidy" to be accepted. You can use https://metacpan.org/pod/distribution/App-githook-perltidy/bin/githook-perltidy#githook-perltidy-install-[-force,-f] to do this automatically and locally for this repo.

t/more-bundle.t
jhthorsen: Thanks for not giving up! The output is completely different from what I can remember, and I think I'll consider using subtests more places 👍

jhthorsen:
Sorry, I missed out on the tidying. Thanks for the improved tests 👍

PR #163, opened 2019-06-11
Bugfix: use time local

augensalat:
### Summary

Date/time checks in `JSON::Validator::Formats` are failing because `Time::Local` is not loaded.

### Motivation
Because it must not fail.

### References
I added a min-test for demonstration.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -15,7 +15,6 @@ use Mojo::Loader;
 use Mojo::URL;
 use Mojo::Util qw(url_unescape sha1_sum);
 use Scalar::Util qw(blessed refaddr);
-use Time::Local ();
 
 use constant CASE_TOLERANT     => File::Spec->case_tolerant;
 use constant COLORS            => eval { require Term::ANSIColor };
--- a/lib/JSON/Validator/Formats.pm
+++ b/lib/JSON/Validator/Formats.pm
@@ -1,6 +1,8 @@
 package JSON::Validator::Formats;
 use Mojo::Base -strict;
 
+require Time::Local;
+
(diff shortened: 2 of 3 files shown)


jhthorsen:
I don't understand what you're saying. I run `prove -vl t/jv-formats.t` and it works nicely. Can you explain how to reproduce the problem?

augensalat:
> I don't understand what you're saying. I run `prove -vl t/jv-formats.t` and it works nicely. Can you explain how to reproduce the problem?

`t/jv-formats.t` does not test `JSON::Validator::Formats` - as none of the other tests does.

Checkout the first commit of this branch and run the test I wrote for `JSON::Validator::Formats`:
```
git checkout 28fe085
prove -vl t/formats.t
```

Or straight to the point:
```
git checkout master
perl -Ilib -MJSON::Validator::Formats -wE 'say JSON::Validator::Formats::check_date("2019-06-12")'
```

jhthorsen:
So you're using JSON::Validator::Formats, without using JSON::Validator?

augensalat:
> So you're using JSON::Validator::Formats, without using JSON::Validator?

Does your question imply that `J:V:Formats` is an internal module? Then you should remove the documentation, or at least include a big DONT-USE-notice.

My answer is: Yes and No.

Yes, because actually I wanted to report another issue in this module (that affects `JSON::Validator`), and then found out, that it is even more broken.

And No, because I do not use it anywhere directly.

Whatever. The `use Time::Local` is in the wrong file. I just want this to be fixed before I report the actual problem.

augensalat:
Two weeks passed - nothing happened.
Is there a chance to make progress on the subject?

PR #165, opened 2019-08-05
Clean up deprecated internal call

mattias-p:
### Summary
Replaces an instance of deprecated API usage with its proper equivalent.

### Motivation
When you call JSON::Validator::Joi::validate you get a deprecation warning from another part of JSON::Validator. That's not right.


jhthorsen:
Oh! I'll merge this at once 👍

PR #166, opened 2019-08-07
fix copy required

sklukin:
Hello 
required dont copy

```
   my $joi = joi->object->props(
        some1 => joi->string->enum([qw/asf fgfg/])->required,
        some2 => joi->string->min(1)->required,
    );
    my $joi_recipient = joi->object->props(
        some3 => joi->object->props(
            some4 => joi->string->min(1)->required,  
            some5 => joi->string->enum([qw/asfd dfdf/]->required
        )
    );
    $joi = $joi->extend( $joi_recipient );
```

required was in first $joi dont by in finish $joi 

my $new_joi = $joi->extend($joi);
Will extend $joi with the definitions in $joi and return a new object.

It is not clear which will be expanded - $jio or $jio ))

my $new_joi = $joi->extend($joi_new);
I think that $joi_new added to $joi and $joi_new remained the same as before

I added the second to the first, and the second remained the same. Not this way?

rebless $self in extend. Why?

### DIFF
--- a/lib/JSON/Validator/Joi.pm
+++ b/lib/JSON/Validator/Joi.pm
@@ -48,7 +48,7 @@ sub extend {
   }
 
   push @{ $clone->{required} }, @{ $self->{required} }
-    if $self->{required} && @{ $self->{required} };
+    if $self->{required} && @{$self->{required}};
 
   return $clone;
 }
@@ -73,6 +73,8 @@ sub props {
     push @{$self->{required}}, $name if $property->{required};
     $self->{properties}{$name} = $property->compile;
   }
+  @{$self->{required}} = sort @{$self->{required}}
+    if $self->{required} && @{$self->{required}};
 
   return $self;
 }
--- a/lib/JSON/Validator/Joi.pm
+++ b/lib/JSON/Validator/Joi.pm
@@ -47,6 +47,9 @@ sub extend {
       for keys %{$self->{properties} || {}};
   }
 
+  push @{ $clone->{required} }, @{ $self->{required} }
+    if $self->{required} && @{ $self->{required} };
+
   return $clone;
 }
 
--- a/t/joi.t
+++ b/t/joi.t
@@ -119,6 +119,36 @@ is_deeply(
 );
 
 is_deeply(
+  edj(
+    joi->object->props(x => joi->integer, y => joi->integer->required)
+      ->extend(
+        joi->object->props(
+            k => joi->object->props(
+                x => joi->string->required,
+                y => joi->string->required
+            )
+        )
+     )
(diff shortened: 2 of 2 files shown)


lib/JSON/Validator/Joi.pm
```diff
@@ -47,6 +47,9 @@ sub extend {
       for keys %{$self->{properties} || {}};
   }
 
+  push @{ $clone->{required} }, @{ $self->{required} }
+    if $self->{required} && @{$self->{required}};
```
jhthorsen: I think this should be unshift() and not push().

I also think this only applies if it's an "object".

And the code does not follow the styling in the rest of the file.

lib/JSON/Validator/Joi.pm
```diff
@@ -70,6 +73,8 @@ sub props {
     push @{$self->{required}}, $name if $property->{required};
     $self->{properties}{$name} = $property->compile;
   }
+  @{$self->{required}} = sort @{$self->{required}}
+    if $self->{required} && @{$self->{required}};
```
jhthorsen: I'm not sure why you want to sort the keys. Is it to get the test through?

t/joi.t
```diff
@@ -118,6 +118,36 @@ is_deeply(
...
+        properties => {
+          x => {type => 'string'}, y => {type => 'string'},
+        },
+        required => ['x', 'y']
+      }
+    },
+    required   => ['k','y'],
+  },
+  'extended object',
+);
```
jhthorsen: This test is not testing what I think you intend it to test.

lib/JSON/Validator/Joi.pm
sklukin: test was failed because` each %%properties`

t/joi.t
sklukin: hmm 
if comment ```
push @{ $clone->{required} }, @{ $self->{required} }
    if $self->{required} && @{$self->{required}};
```
test fail

sklukin: bag... got it )) thanks

lib/JSON/Validator/Joi.pm
sklukin: bag... got it )) thanks

lib/JSON/Validator/Joi.pm
sklukin: Ok )

t/joi.t
jhthorsen: Yes, but your structure has a nested object, which is not touched by the merging. extend() only works on the first level.

sklukin:
Thanks very much ))

PR #173, opened 2019-09-01
new way to obfuscate file path

elcamlost:
### Summary
Improve readability of doc page generated by Mojolicious::Plugin::OpenAPI

### Motivation
Current way leads to unreadable definition names and doesn't hide filenames.

Compare 
before
and
after


jhthorsen:
This PR is invalid because it has no tests and even breaks some of the existing tests.

PR #177, opened 2019-10-18
Add support for changing definitions path when working with $ref

HEM42:
### Summary

Allow for definitions path to be changed, today the default behavior for adding refs into `/definitions/some_def` will render a OpenAPI v3 spec wrongly.

a $ref like `#/components/schemas/myschema` will be rendered as `/definitions/components_schemas_myschema`, which is not valid according the OpenAPI v3 spec

the `generate_definitions_path` function makes it possible to override this behavior, keeping JSON::Validator neutral to the proposed change.

These changes is based on the generate_definitions_path branch, but with a fix for the test

### Motivation

OpenAPIv3 has chosen to rearrange its definitions, into a deeper path as an example `/components/schemas/myschema`

This change makes it possible for modules like Mojolicious::Plugin::OpenAPI to support the v3, with minimal effort

Earlier attempts to make this right eq. #174 was not general enough

Hope that this time around the change is general enough to be accepted 
 
### References

* https://swagger.io/specification/
* Old PR, #174 which also have bit of information on generate_definitions_path 
* PR #175 is not needed if this PR is accepted
* jhthorsen/mojolicious-plugin-openapi/issues/138

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -102,7 +102,7 @@ sub bundle {
       unless ($self->{seen_ref}{$tied->fqn}++) {
         push @topics,
           [_node($schema, $path, 1, 0) || {}, _node($bundle, $path, 1, 1)];
-        push @topics, [$tied->schema, $bundle->{$path->[0]}{$path->[1]} ||= {}];
+        push @topics, [$tied->schema, _node($bundle, $path, 0, 1)];
       }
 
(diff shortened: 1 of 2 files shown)


jhthorsen:
I'm not too fond of merge commits in PR's, so I pushed the changes directly in 98dbf69ba67b156f6a0371a585cdd586a97d1d79

Thanks for fixing the unit test 👍

Issue #170, opened 2019-08-20
more-bundle.t hangs on windows


chorny:
* JSON::Validator version: 3.14
* Perl version: 5.26.0 and others (probably depending on YAML::XS 0.67 availability)
* Operating system: Windows XP

```
Deep recursion on subroutine "JSON::Validator::_resolve" at C:\Strawberry260\cpan\build\JSON-Validator-3.14-0\blib\lib/JSON/Validator.pm line 542.
Deep recursion on subroutine "JSON::Validator::_resolve_ref" at C:\Strawberry260\cpan\build\JSON-Validator-3.14-0\blib\lib/JSON/Validator.pm line 505.
YAML::XS Error: Expected STREAM_START_EVENT; Got: 0 != 1 at C:\Strawberry260\cpan\build\JSON-Validator-3.14-0\blib\lib/JSON/Validator.pm line 358.
(hangs)
```

perlpunk:
@chorny  Just to factor out YAML::XS here - could you possibly test with the newest YAML::XS 0.80?

The error message could point to a memory leak when loading invalid YAML, and I just fixed a leak.

chorny:
Tested with 0.80, same problem

jhthorsen:
Sorry, I don't have windows, so I won't be able to figure out why this doesn't work.

Also seems to only be an issue on certain versions of Windows..? https://app.thorsen.pm/cpantestreport/dist/JSON-Validator-3.14?osname=MSWin32

karenetheridge:
Well that's indicating that a change made in `bundle` (or one of the private methods it uses) didn't work, since the tests in more-bundle.t *used* to cause infinite recursion until the implementation was changed.  Interesting that this only happens on windows! perhaps a difference with how Tie::RefHash is implemented?

sudo-batman:
Just FYI - I also got the similar issue with the new version.

  JSON::Validator version: 3.16
  Perl version: 5.30.0 (Strawberry)
  Operating system: Windows 10

```
t/load-yaml.t .......................... ok
t/more-bundle.t ........................ Deep recursion on subroutine "JSON::Validator::_resolve" at C:\berrybrew\5.30.0_64\cpan\build\JSON-Validator-3.16-2\blib\lib/JSON/Validator.pm line 574.
[...]

PR #178, opened 2019-10-24
fix for infinite recursion on Windows, issue #170

chorny:
So it would be lowercase and match

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -524,7 +524,9 @@ sub _location_to_abs {
   # definitely relative now
   if ($base->isa('Mojo::File')) {
     return $base if !length $location;
-    return $base->sibling(split '/', $location)->realpath;
+    my $tpath = $base->sibling(split '/', $location)->realpath;
+    #lc fixes https://github.com/mojolicious/json-validator/issues/170
+    return CASE_TOLERANT ? lc($tpath) : $tpath;
   }
   return $location_as_url->to_abs($base);
 }


lib/JSON/Validator.pm
```diff
@@ -524,7 +524,9 @@ sub _location_to_abs {
   # definitely relative now
   if ($base->isa('Mojo::File')) {
     return $base if !length $location;
-    return $base->sibling(split '/', $location)->realpath;
+    my $tpath = $base->sibling(split '/', $location)->realpath;
```
jhthorsen: What is "t" in "tpath"?

lib/JSON/Validator.pm
```diff
@@ -524,7 +524,9 @@ sub _location_to_abs {
...
+    #lc fixes https://github.com/mojolicious/json-validator/issues/170
```
jhthorsen: This comment is not needed.

karenetheridge:
So, it appears (from inferring from the patch provided) that infinite recursion is occuring because two paths are being compared case-sensitively on a case-insensitive filesystem, and they should match but they do not. Where in the tests is that happening? That is, where are two paths provided that are identical-except-for-case?

This also provides a hint for writing a demonstration test -- one could perhaps hack CASE_TOLERANT so it is the same value across all platforms (for this one test), and demonstrate how behaviour is different with and without the `lc`.

I'm also curious where the path comparison is taking place, such that lower-casing is necessary -- because `lc` is not necessarily the right thing to do here, given various locales' handling of case.  When comparing strings, foldcase (`fc`) is preferred.

jhthorsen:
I'm going to close this, since it seems like @charsbar is not interested in this anymore.

charsbar:
@jhthorsen , well maybe it was @chorny not me, but anyway, I sent a modified version of this PR #180

jhthorsen:
That was a bad autocompleted from me. Sorry about that 😢

Issue #102, opened 2018-03-12
bundle.t hangs on Windows


chorny:
Strawberry perl 5.26.0 32-bit + 64-bit int.

```
# [1] replace=1
ok 1 - [1] replace=1
# [1] replace=0
ok 2 - [1] name still in definitions
ok 3 - [1] added to definitions
ok 4 - [1] new age ref
ok 5 - [1] same name ref
ok 6 - [1] age $ref point to /definitions/b_json__definitions_age
ok 7 - [1] name $ref point to /definitions/name
# [2] replace=1
ok 8 - [2] replace=1
# [2] replace=0
ok 9 - [2] name still in definitions
ok 10 - [2] added to definitions
ok 11 - [2] new age ref
ok 12 - [2] same name ref
ok 13 - [2] age $ref point to /definitions/b_json__definitions_age
ok 14 - [2] name $ref point to /definitions/name
# [3] replace=1
ok 15 - [3] replace=1
# [3] replace=0
ok 16 - [3] name still in definitions
ok 17 - [3] added to definitions
ok 18 - [3] new age ref
ok 19 - [3] same name ref
ok 20 - [3] age $ref point to /definitions/b_json__definitions_age
ok 21 - [3] name $ref point to /definitions/name
ok 22 - get /name/$ref
ok 23 - get /name/type
ok 24 - get /name/$ref
ok 25 - schema get /name/type
ok 26 - schema get /name/$ref
ok 27 - no dup definitions
Deep recursion on subroutine "JSON::Validator::_resolve" at C:\Strawberry260\cpan\build\JSON-Validator-2.05-0\blib\lib/JSON/Validator.pm line 432.
Deep recursion on subroutine "JSON::Validator::_resolve_ref" at C:\Strawberry260\cpan\build\JSON-Validator-2.05-0\blib\lib/JSON/Validator.pm line 407.
Deep recursion on subroutine "JSON::Validator::_resolve_ref" at C:\Strawberry260\cpan\build\JSON-Validator-2.05-0\blib\lib/JSON/Validator.pm line 407.
Deep recursion on subroutine "JSON::Validator::_resolve_ref" at C:\Strawberry260\cpan\build\JSON-Validator-2.05-0\blib\lib/JSON/Validator.pm line 407.
(hangs)
```

Also see this cpantesters report: http://www.cpantesters.org/cpan/report/5c245523-6bf6-1014-8ac2-ceba6fcb7d32

vejbaek:
[...]

PR #180, opened 2019-12-28
Fix for deep recursion on Windows #170

charsbar:
# Summary
This is the modified version of #178 by @chorny.

### Motivation
The point of this fix is to use the same type of id as the _load_schema() method returns at [the line 342](https://github.com/mojolicious/json-validator/blob/6ba3427f31a808541afd6fc9b27b04515c2e38f2/lib/JSON/Validator.pm#L342) on a case tolerant file system.

I wanted to add a test that uses MixedCase files but simple mocking of CASE_TOLERANT constant doesn't work under a case intolerant system because _load_schema fails to find the other schema file because of this case lowering. Removing case lowering of the line 342 would also work, but considering of #102, I assume it's better to keep it as is. Maybe we can consider t/more-bundle.t as a required failing test. Some of the Win32 reports say the tests have "pass"ed but that's just because YAML::XS is not installed there and the test is skipped.

### References
#170, #178, #102

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -547,7 +547,8 @@ sub _location_to_abs {
   # definitely relative now
   if ($base->isa('Mojo::File')) {
     return $base if !length $location;
-    return $base->sibling(split '/', $location)->realpath;
+    my $path = $base->sibling(split '/', $location)->realpath;
+    return CASE_TOLERANT ? lc($path) : $path;
   }
   return $location_as_url->to_abs($base);
 }


jhthorsen:
Can’t you just do something like this in the test?

plan skip_all => 'not case tolerant' unless JSON::Validator::CASE_TOLERANT;

charsbar:
Skipping a test doen't fix anything actually, and people can install this module without testing already (if they really need this, like me).

I'm not sure why you suggest skipping, but if you need to be assured before merging, GitHub Actions or AppVeyor may help.

jhthorsen:
CPAN testers will still run the test, and let me know in case I remove that logic in the future. And the test will be run for anyone who it’s relevant for.

charsbar:
You've added the CASE_TOLERANT trick for Windows (#102), instead of skipping t/bundle.t. And then t/more-bundle.t added recently revealed the previous fix was insufficient. I believe it's more appropriate to fix it again.

charsbar:
Sorry, I pushed the wrong button.

jhthorsen:
Oh! I haven’t noticed that CPAN testers have failing tests on windows.

charsbar:
Thank you!

PR #183, opened 2020-01-29
fix reported error

karenetheridge:
I just noticed this when merging changes (sorry!). It is very low priority as JSON_VALIDATOR_REPORT isn't even documented.

(I also did a perltidy.)


jhthorsen:
JSON_VALIDATOR_REPORT is only useful for developers trying to figure out weird bugs.

Thanks for the fix :)

PR #184, opened 2020-02-06
Add clearer error when more than one oneOf rule matched

### LINKED ISSUE #164: handling of 'oneOf' rules less than optimal
* JSON::Validator version: 3.11
* Perl version: not relevant

When more than one 'oneOf' rule matches, the error is not correct. The following script demonstrates the issue:

```
use strict;
use warnings;
use JSON::Validator;

my $schema = {
    type => 'object',
    properties => {},
    additionalProperties => {
        oneOf => [
            { type => 'integer', minimum => 1 },
            { type => 'integer', minimum => 2 },
            { type => 'integer', maximum => 3 },
        ],
    },
};

my $jv = JSON::Validator->new;
$jv->load_and_validate_schema($schema, { schema => 'http://json-schema.org/draft-07/schema#' });

my $data = { foo => 1 };
my @errors = $jv->validate($data, $schema);
use Data::Dumper;
print STDERR "### got errors: ", Dumper(\@errors);

# got message: '/oneOf/1 1 < minimum(2)'
# but it should be: 'oneOf Multiple rules matched: /oneOf/0, /oneof/2.'
```

In the `_validate_one_of` sub, I think we should be tracking all the rules that *do* match, and then if `@matched` != 1, the path to all the matching rules can be easily provided in the error message.  As it is now, only the *failing* rules are tracked, and this is used to construct the error message, but that is not helpful here.

karenetheridge:
fixes #164.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -731,9 +731,9 @@ sub _validate_one_of {
   $self->_report_schema($path, 'oneOf', $rules) if REPORT;
   local $self->{grouped} = $self->{grouped} + 1;
 
-  my $i = 0;
+  my ($i, @passed) = (0);
   for my $rule (@$rules) {
-    my @e           = $self->_validate($_[1], $path, $rule) or next;
+    my @e = $self->_validate($_[1], $path, $rule) or push @passed, $i and next;
     my $schema_type = _guess_schema_type($rule);
     push @errors, [$i, @e] and next if !$schema_type or $schema_type eq $type;
     push @expected, $schema_type;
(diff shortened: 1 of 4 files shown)


karenetheridge:
> Would it be possible to let the user know which rules that failed?

Yes, that should be possible with a bit of juggling. I'll push an update.

jhthorsen:
Don’t do it if it’s a lot of hassle/lines of code.

karenetheridge:
It would be a few lines of code -- we can get all the failure rules numbers by doing

    -    push @expected, $schema_type;
    +    push @expected, [$i, $schema_type];

and then `@failed = sort {$a <=> $b} map $_->[0], @errors, @expected`, but then we need to subtract that list from `0..@rules-1` to get the rules that *passed*. I don't see how to do it without at least one extra local variable :)

karenetheridge:
I pushed another commit that uses the indexes of the matching oneOf rules, which you can squash if you like it, or ignore if you don't :)

karenetheridge:
I force-pushed a much nicer change for adding passing indexes. :)

karenetheridge:
and done!

PR #185, opened 2020-02-07
Split JSON::Validator into subclasses for more flexibility

jhthorsen:
JSON::Validator (today) has a lot of code. 1591 lines to be exact - including documentation. I think this makes it hard to navigate the codebase and figure out how it is all connected. Another issue is that all the "_validate_x" methods has a mix of all the different draft version, which might make it unnecessarily slow.

I propose splitting the validation methods into sub-classes, such as JSON::Validator::Draft4, ::Draft6, ..., but starting out with ::DraftX, which will have the current functionality.

In addition, this PR adds the JSON::Validator::Util package with various utility functions. I'm splitting this out, since they might be useful to include in the different ::Draft sub-classes and/or useful for end users.

This PR cannot be merged yet, because...
* M::P::OpenAPI needs to be updated.
* Need to write documentation for ::Util and ::DraftX.


jhthorsen:
Going to close this, since it might not be the best way to go. Will probably just add the ::Draft classes directly in master instead.

PR #190, opened 2020-02-11
fix use of true, false (sub)schemas

karenetheridge:
### Summary
Now accepts (sub)schemas that are the boolean values `true` or `false`.

### Motivation
`true` and `false` are valid subschemas in draft 6 and later, as per https://json-schema.org/understanding-json-schema/basics.html

### DIFF
--- a/Changes
+++ b/Changes
@@ -47,7 +47,7 @@ Revision history for perl distribution JSON-Validator
 3.09 2019-05-04T22:28:55+0700
  - Prettier definition names from bundle().
  - Changed default bundle() definitions location from "x-bundle" to "definitions".
- - Deprecated bundle({ref_ref => ...})
+ - Deprecated bundle({ref_key => ...})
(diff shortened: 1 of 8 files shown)


karenetheridge:
> JSON::PP::true, JSON::PP::false can be replaced with just false and true.

fixed.

t/util.t
```diff
@@ -28,6 +28,7 @@ ok is_type(42,    'NUM'),             'is_type 42';
 ok is_type(false, 'BOOL'),            'is_type BOOL';
 ok !is_type('2', 'NUM'),  'is_type 2';
 ok !is_type(0,   'BOOL'), 'is_type BOOL';
+ok is_type($v, 'HASH'), 'blessed object is_type HASH';
```
jhthorsen: Isn't this the same test as on line 23?

karenetheridge: lol yes. I'll remove the commit.

PR #195, opened 2020-02-13
misc small changes from ether

karenetheridge:
Some small fixes to docs and style that I made locally while stepping through code while debugging...  these should be pretty uncontroversial changes I think. :)

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -1039,7 +1039,7 @@ Here are some resources that are related to JSON schemas and validation:
 
 =item * L<http://json-schema.org/documentation.html>
 
-=item * L<http://spacetelescope.github.io/understanding-json-schema/index.html>
+=item * L<https://json-schema.org/understanding-json-schema/index.html>
 
 =item * L<https://github.com/json-schema/json-schema/>
 
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -112,7 +112,7 @@ sub bundle {
       }
     }
     elsif (ref $from eq 'HASH') {
-      while (my ($key, $value) = each %$from) {
+      for my $key (keys %$from) {
         $to->{$key} //= $cloner->($from->{$key});
       }
     }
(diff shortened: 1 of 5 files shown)


lib/JSON/Validator.pm
```diff
@@ -300,22 +300,12 @@ sub _load_schema_from_text {
   return Mojo::JSON::decode_json($$text) if $$text =~ /^\s*\{/s;
 
   # YAML
-  $visit = sub {
```
jhthorsen: You can remote `my $visit` in this sub as well.

lib/JSON/Validator.pm
```diff
@@ -875,10 +864,11 @@ sub _validate_type_object {
       : $self->_validate($data, $path, $schema->{then} // {});
   }
 
-  my $coerce = $self->{coerce}{defaults};
-  while (my ($k, $r) = each %{$schema->{properties}}) {
+  my %rules;
+  for my $k (keys %{$schema->{properties}}) {
```
jhthorsen: Why did you change from `while` to `for` here?

lib/JSON/Validator.pm
```diff
@@ -887,11 +877,12 @@ sub _validate_type_object {
     }
   }
 
-  while (my ($p, $r) = each %{$schema->{patternProperties} || {}}) {
+  for my $p (keys %{$schema->{patternProperties} || {}}) {
```
jhthorsen: Why did you change from `while` to `for` here?

lib/JSON/Validator.pm
```diff
@@ -904,7 +895,7 @@ sub _validate_type_object {
     return E $path, [object => additionalProperties => join '/', @k];
   }
 
-  for my $k (sort keys %required) {
+  for my $k (sort uniq @{$schema->{required} || []}) {
```

lib/JSON/Validator.pm
karenetheridge: I explained that in the comment for commit  46c2b23.

jhthorsen:
I'll merge it even though I don't fully agree on the while/for changes. (Have to make a new release)

Issue #196, opened 2020-02-13
all of allOf, anyOf, oneOf can exist together in a schema


karenetheridge:
I noticed that the `_validate` sub skips anyOf and oneOf rules if allOf was present. But nowhere in the spec does it say that these can't appear together. We should run them all and collect all the errors, even if it's nonsensical to combine the rules.

references:
https://json-schema.org/understanding-json-schema/reference/combining.html
https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.7

karenetheridge:
(fixed in a branch, not yet PR'd.)

Issue #192, opened 2020-02-12
unsupported keyword for object type: dependencies


karenetheridge:
described here: 
https://json-schema.org/understanding-json-schema/reference/object.html?highlight=dependencies
https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.5.7

PR #197, opened 2020-02-14
fix if-then-else, anyOf, dependencies

karenetheridge:
### Summary
fixes bugs: #190, #196, #192 where various keywords were not fully supported

### Motivation
see their respective issues for background and references

I am not quite sure what error messages are best: "Missing property" was already being used by the 'required' keyword, and it's not great to have two different errors sharing the same message, so I made the 'required' error "Missing required property", so 'dependencies' can have "Missing dependent property". However, changing an existing error message might break unit tests in downstream code.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -54,7 +54,7 @@ has ua => sub {
 
 sub bundle {
   my ($self, $args) = @_;
-  my ($cloner, $tied);
+  my ($cloner);
 
   my $schema
     = $args->{schema} ? $self->_resolve($args->{schema}) : $self->schema->data;
@@ -64,6 +64,7 @@ sub bundle {
     $cloner = sub {
       my $from      = shift;
       my $from_type = ref $from;
+      my $tied;
       $from = $tied->schema if $from_type eq 'HASH' and $tied = tied %$from;
       my $to = $from_type eq 'ARRAY' ? [] : $from_type eq 'HASH' ? {} : $from;
       push @topics, [$from, $to] if $from_type;
@@ -75,7 +76,7 @@ sub bundle {
       my $from      = shift;
       my $from_type = ref $from;
 
-      $tied = $from_type eq 'HASH' && tied %$from;
+      my $tied = $from_type eq 'HASH' && tied %$from;
       unless ($tied) {
         my $to = $from_type eq 'ARRAY' ? [] : $from_type eq 'HASH' ? {} : $from;
         push @topics, [$from, $to] if $from_type;
@@ -293,7 +294,6 @@ sub _load_schema {
 
 sub _load_schema_from_text {
   my ($self, $text) = @_;
-  my $visit;
 
   # JSON
   return Mojo::JSON::decode_json($$text) if $$text =~ /^\s*\{/s;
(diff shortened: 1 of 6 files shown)


lib/JSON/Validator/Error.pm
```diff
@@ -34,7 +34,8 @@ our $MESSAGES = {
     additionalProperties => 'Properties not allowed: %3.',
     maxProperties        => 'Too many properties: %3/%4.',
     minProperties        => 'Not enough properties: %3/%4.',
-    required             => 'Missing property.',
+    required             => 'Missing required property.',
+    dependencies         => 'Missing dependent property.',
```
jhthorsen: I think this is incomplete: It should mention the path/name of the dependent property. How about something like "%3 missing dependent property." or something? If you can make it shorter or rephrase it so "%3" comes at the end, then that would be super.

lib/JSON/Validator.pm
```diff
@@ -551,15 +549,25 @@ sub _validate {
...
   }
-  elsif ($rules = $schema->{anyOf}) {
+
+  if (my $rules = $schema->{anyOf}) {
     push @errors,
       $self->_validate_any_of($to_json ? $$to_json : $_[1], $path, $rules);
   }
-  elsif ($rules = $schema->{oneOf}) {
+
+  if (my $rules = $schema->{oneOf}) {
```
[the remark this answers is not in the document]
karenetheridge: a combination of oneOf, anyOf and allOf all at the same time? hey, you never know :D

lib/JSON/Validator.pm
```diff
@@ -900,6 +902,22 @@ sub _validate_type_object {
...
+        push @errors, E json_pointer($path, $dep_key),
+          [object => dependencies => $data_key];
+      }
+    }
+    elsif (ref $schema->{dependencies}{$data_key} eq 'HASH') {
+      push @errors,
+        $self->_validate_type_object($data, $path,
+        $schema->{dependencies}{$data_key});
+    }
+  }
```
jhthorsen: Sorry for going back/forward on this code, but I think this is easier to read:

```perl
  my $dependencies = $schema->{dependencies} || {};
  for my $k (keys %$dependencies) {
    next if not exists $data->{$k};

    if (ref $dependencies->{$k} eq 'ARRAY') {
      push @errors,
        map { E json_pointer($path, $_), [object => dependencies => $k]; }
        grep { !exists $data->{$_} } @{$dependencies->{$k}};
    }
    elsif (ref $dependencies->{$k} eq 'HASH') {
      push @errors,
        $self->_validate_type_object($data, $path, $dependencies->{$k});
    }
  }
```

lib/JSON/Validator/Error.pm
```diff
@@ -35,6 +35,7 @@ our $MESSAGES = {
     maxProperties        => 'Too many properties: %3/%4.',
     minProperties        => 'Not enough properties: %3/%4.',
     required             => 'Missing property.',
+    dependencies         => 'Missing property: Dependee: %3.',
```
jhthorsen: Too many colons. How about "Missing property. Dependee: %3."

lib/JSON/Validator/Error.pm
jhthorsen: See https://github.com/mojolicious/json-validator/pull/197#discussion_r379681124

lib/JSON/Validator/Error.pm
karenetheridge: fixed.

lib/JSON/Validator.pm
karenetheridge: the map and grep is not easier to read.

karenetheridge: ..but ok.

jhthorsen:
Awesome 👍

Issue #198, opened 2020-02-17
Order of missing required fields changed in Version 3.21


perlpunk:
* JSON::Validator version: 3.21, 3.22
* Perl version: 5.26
* Operating system: openSUSE 15.1

### Steps to reproduce the behavior
I don't have a code example yet, but the order of missing required fields changed. Before they were sorted.

### Expected behavior
Missing fields should be reported in sorted order.

### Actual behavior
Missing fields are reported in apparently random order.

I think the culprit is this line:
https://github.com/mojolicious/json-validator/blob/master/lib/JSON/Validator.pm#L899
```
for my $k (sort uniq @{$schema->{required} || []}) {
```
Because `sort` is special, I think it is taking the `uniq` function as the sorting function. This can be fixed by using parens:
```
for my $k (sort (uniq( @{$schema->{required} || []} ) )) {
```

karenetheridge:
> Missing fields should be reported in sorted order.

That's not in the spec though :)  If we were to have any sort of order, I think probably the best order is the order that the fields are listed in in the `required` key.

> Because sort is special, I think it is taking the uniq function as the sorting function

Aha!  it can be easily fixed by switching the order: `uniq sort @list`

karenetheridge:
I just noticed that one problem in using a home-grown `uniq` sub, instead of List::Util::uniq (why, anyway? List::Util is in core!) is that it loses the original ordering of the list. :(

jhthorsen:
I wasn't able to figure out when List::Util::uniq became part of core. Can you please tell me which version of Perl that is? I know List::Util have been there for "ages", but the uniq() functions has not been there that long...

jhthorsen:
Oh! Just learned about `corelist -a` 😄 So it was added in v5.25.1. That's relatively recent, but old enough for me to add List::Util to `Makefile.PL`.

perlpunk:
Thanks, explicitly defining the sort function via `{ $a cmp $b }` also fixes the issue.

Not sure why you say `JSON::Validator::Util::uniq` is losing the the original order, @karenetheridge . To me it looks like it works correctly, just that the call of `sort` was wrong.

perlpunk:
> Just learned about `corelist -a`

See also http://perlpunks.de/corelist :)

perlpunk:
> > Missing fields should be reported in sorted order.
> 
> That's not in the spec though :) If we were to have any sort of order, I think probably the best order is the order that the fields are listed in in the `required` key.

[...]

PR #199, opened 2020-02-18
fix oneOf with type => [array]

karenetheridge:
### Summary
fixes more edge cases with oneOf

### Motivation
Many of the new tests I added will fail without the accompanying code changes.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -519,8 +519,7 @@ sub _validate {
   # Test base schema before allOf, anyOf or oneOf
   if (ref $type eq 'ARRAY') {
     push @{$self->{temp_schema}}, [map { +{%$schema, type => $_} } @$type];
-    push @errors,
-      $self->_validate_any_of($to_json ? $$to_json : $_[1],
+    return $self->_validate_any_of_types($to_json ? $$to_json : $_[1],
       $path, $self->{temp_schema}[-1]);
   }
   elsif ($type) {
@@ -574,69 +573,91 @@ sub _validate {
 
 sub _validate_all_of {
   my ($self, $data, $path, $rules) = @_;
-  my $type = data_type $data, $rules;
-  my (@errors, @expected);
+  my @errors;
 
   my $i = 0;
   for my $rule (@$rules) {
     next unless my @e = $self->_validate($_[1], $path, $rule);
-    my $schema_type = schema_type $rule;
-    push @expected, $schema_type if $schema_type;
-    push @errors, [$i, @e] if !$schema_type or $schema_type eq $type;
+    push @errors, [$i, @e];
   }
   continue {
     $i++;
   }
 
-  return E $path, [allOf => type => join('/', uniq @expected), $type]
-    if !@errors and @expected;
-  return prefix_errors allOf => @errors if @errors;
-  return;
+  return if not @errors;
+
+  return prefix_errors(allOf => @errors)
+    if @errors == 1
+    or grep $_->[1]->details->[1] ne 'type', @errors;
+
+  # combine all 'type' errors together
+  my $want_types = join('/', uniq map $_->[1]->details->[0], @errors);
+  return E $path, [allOf => type => $want_types, $errors[-1][1]->details->[2]];
+}
+
+sub _validate_any_of_types {
+  my ($self, $data, $path, $rules) = @_;
+  my @errors;
+
+  for my $rule (@$rules) {
+    return unless my @e = $self->_validate($_[1], $path, $rule);
+    push @errors, @e;
+  }
+
+ # disregard any 'type' errors in favour of a more specific error, when possible
+  if (my @e = grep $_->details->[1] ne 'type', @errors) {
+    return @e;
+  }
+
+  # combine all 'type' errors together
+  my $want_types = join('/', uniq map $_->details->[0], @errors);
+  return E $path, [$want_types => 'type', $errors[-1]->details->[2]];
 }
 
 sub _validate_any_of {
   my ($self, $data, $path, $rules) = @_;
-  my $type = data_type $data, $rules;
-  my (@e, @errors, @expected);
+  my @errors;
 
   my $i = 0;
   for my $rule (@$rules) {
-    @e = $self->_validate($_[1], $path, $rule);
-    return unless @e;
-    my $schema_type = schema_type $rule;
-    push @errors, [$i, @e] and next if !$schema_type or $schema_type eq $type;
-    push @expected, $schema_type;
+    return unless my @e = $self->_validate($_[1], $path, $rule);
+    push @errors, [$i, @e];
   }
   continue {
     $i++;
   }
 
-  my $expected = join '/', uniq @expected;
-  return E $path, [anyOf => type => $expected, $type] unless @errors;
-  return prefix_errors anyOf => @errors;
+  return prefix_errors(anyOf => @errors)
+    if grep $_->[1]->details->[1] ne 'type', @errors;
+
+  # combine all 'type' errors together
+  my $want_types = join('/', uniq map $_->[1]->details->[0], @errors);
+  return E $path, [anyOf => type => $want_types, $errors[-1][1]->details->[2]];
 }
 
 sub _validate_one_of {
   my ($self, $data, $path, $rules) = @_;
-  my $type = data_type $data, $rules;
-  my (@errors, @expected);
+  my @errors;
 
   my ($i, @passed) = (0);
   for my $rule (@$rules) {
     my @e = $self->_validate($_[1], $path, $rule) or push @passed, $i and next;
-    my $schema_type = schema_type $rule;
-    push @errors, [$i, @e] and next if !$schema_type or $schema_type eq $type;
-    push @expected, $schema_type;
+    push @errors, [$i, @e];
   }
   continue {
     $i++;
   }
 
   return if @passed == 1;
-  return E $path, [oneOf => 'all_rules_match'] unless @errors + @expected;
+  return E $path, [oneOf => 'all_rules_match'] unless @errors;
   return E $path, [oneOf => 'n_rules_match', join(', ', @passed)] if @passed;
-  return prefix_errors oneOf => @errors if @errors;
-  return E $path, [oneOf => type => join('/', uniq @expected), $type];
+
+  return prefix_errors(oneOf => @errors)
+    if grep $_->[1]->details->[1] ne 'type', @errors;
+
+  # combine all 'type' errors together
+  my $want_types = join('/', uniq map $_->[1]->details->[0], @errors);
+  return E $path, [oneOf => type => $want_types, $errors[-1][1]->details->[2]];
 }
 
 sub _validate_number_max {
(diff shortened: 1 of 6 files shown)


lib/JSON/Validator.pm
```diff
@@ -573,69 +573,85 @@ sub _validate {
...
+
+  return prefix_errors allOf => @errors;
 }
 
 sub _validate_any_of {
-  my ($self, $data, $path, $rules) = @_;
-  my $type = data_type $data, $rules;
-  my (@e, @errors, @expected);
+  my ($self, $data, $path, $rules, $prefix) = @_;
+  $prefix //= 1;
```
jhthorsen: I don't like this `$prefix` variable. It's a bit worse than my `$expected` variable in `_validate_type_number()` 😥 Can we make prefix_errors() smarter instead?

lib/JSON/Validator.pm
```diff
@@ -573,69 +573,85 @@ sub _validate {
...
+  my (@e, @errors);
 
   my $i = 0;
   for my $rule (@$rules) {
     @e = $self->_validate($_[1], $path, $rule);
+
+    # TODO: the spec might not think that short-circuiting is ok here
```
jhthorsen: What do you mean with this comment? Why would you use CPU cycles checking for more errors, if it already fulfils anyOf?

lib/JSON/Validator.pm
```diff
@@ -573,69 +573,85 @@ sub _validate {
...
+  if (grep $_->[1]->details->[1] ne 'type', @errors) {
+    return $prefix ? prefix_errors(anyOf => @errors) : (map $_->[1], @errors);
+  }
+
+  my $want_types = join('/', uniq map $_->[1]->details->[0], @errors);
+  return E $path,
+    [
+    ($prefix ? (anyOf => type => $want_types) : ($want_types => 'type')),
+    $errors[-1][1]->details->[2]
+    ];
```
jhthorsen: Pretty sure this can be made short as well. See my comment above about "This is very difficult to read". (I'm thinking "extract `@details` first) This goes for `_validate_one_of()` below as well.

t/issue-103-one-of.t
```diff
@@ -5,7 +5,8 @@ validate_ok {who_id => 'WHO', expire => '2018-01-01', amount => 1000,
   desc => 'foo'}, 'data://main/example.json',
   E('/sym',      '/oneOf/0/allOf/0/allOf/0 Missing property.'),
   E('/template', '/oneOf/0/allOf/2 Missing property.'),
-  E('/sym',      '/oneOf/1/allOf/0 Missing property.');
+  E('/sym',      '/oneOf/1/allOf/0 Missing property.'),
+  E('/',         '/oneOf/2 Expected string - got object.');
```
jhthorsen: 👍

t/jv-allof.t
```diff
@@ -39,4 +39,11 @@ validate_ok 'he110th3re', $schema,
   E('/', '/oneOf/1 String is too long: 10/4.'),
   E('/', '/oneOf/1 String does not match ^[a-z]+$.');
 
+
```
jhthorsen: Extra newline.

t/jv-allof.t
```diff
@@ -39,4 +39,11 @@ validate_ok 'he110th3re', $schema,
...
+validate_ok 'hello', {type => ['integer', 'boolean']},
+  E('/', 'Expected integer/boolean - got string.');
```
jhthorsen: I think the message should have the types sorted, despite the order in `type`. (See #198)

t/jv-anyof.t
```diff
@@ -63,4 +67,15 @@ validate_ok(
   E('/b', '/anyOf/1 Missing property.'),
 );
 
+validate_ok 'hello', {type => ['integer', 'string'], enum => [123, 'HELLO']},
+  E('/', 'Expected integer - got string.'),
```
jhthorsen: This doesn't sounds right. We got a valid type, didn't we..?

t/jv-oneof.t
```diff
@@ -23,29 +23,50 @@ $schema = {oneOf => [{type => 'object'}, {type => 'string', multipleOf => 3}]};
 validate_ok 13, $schema, E('/', '/oneOf Expected object/string - got number.');
 
 $schema = {oneOf => [{type => 'object'}, {type => 'number', multipleOf => 3}]};
-validate_ok 13, $schema, E('/', '/oneOf/1 Not multiple of 3.');
+validate_ok 13, $schema, E('/', '/oneOf/0 Expected object - got number.'),
+  E('/', '/oneOf/1 Not multiple of 3.');
 
 # Alternative oneOf
-# http://json-schema.org/latest/json-schema-validation.html#anchor79
+# https://json-schema.org/draft-07/json-schema-validation.html#rfc.section.7
```
jhthorsen: 👍

lib/JSON/Validator.pm
karenetheridge: The draft 2019-09 spec has a few things to say about verifying all branches of a conditional even if some parts already gave a sufficient answer to report up to the next level.  I didn't mean to commit that comment; it was a note to myself to go back and read the spec more carefully.

t/jv-allof.t
karenetheridge: I disagree!  Any time there is a defined order in the schema (an arrayref, vs a hashref) I think we should keep the order as presented by the user.

t/jv-oneof.t
karenetheridge: It took a bit of sleuthing to figure out what part of the spec was being referenced there. I only figured it out when I broken the 'format' tests and found the bit in the spec explaining how it worked differently than everything else -- I had altered `schema_type` to infer `string` from `format`, and then found the tests that broke as a consequence.

karenetheridge:
I have created a separate PR for the first change.  That commit is still contained in this PR as the code change is also needed here, but it will rebase away when it is merged to master.

lib/JSON/Validator.pm
karenetheridge: I don't either, but it was late when I got to this point and I kind of gave up. :(  The key here is indicating to _validate_any_of whether it should actually be prefixing a path to the errors it got, or not -- when called via `type => [...]` there isn't any literal anyOf rule from the user, so we should just report the error(s) as-is without a prefix.

karenetheridge:
`@expected` is removed in the _validate_(all|any|one)_of subs because it doesn't actually help anything -- the `_validate` call already did the type check and returned the proper error, and also `schema_type` can't return a useful result in the cases where `type => ..` is a list of types.

We don't actually need to do anything extra at all at the end of these subs if we don't care about joining all the `type` errors into one -- the code gets much simpler if instead of:

    E('/', '/anyOf/1 Expected integer/boolean - got string.');

we can return:

    E('/', '/anyOf/1 Expected integer - got string.');
    E('/', '/anyOf/1 Expected boolean - got string.');

t/jv-allof.t
jhthorsen: Most people don’t care about order (based on experience working with people) when they write their schemas, so we don’t either when creating the error messages.

lib/JSON/Validator.pm
```diff
@@ -519,9 +519,8 @@ sub _validate {
   # Test base schema before allOf, anyOf or oneOf
   if (ref $type eq 'ARRAY') {
     push @{$self->{temp_schema}}, [map { +{%$schema, type => $_} } @$type];
-    push @errors,
-      $self->_validate_any_of($to_json ? $$to_json : $_[1],
-      $path, $self->{temp_schema}[-1]);
+    return $self->_validate_any_of_types($to_json ? $$to_json : $_[1],
+      $path, $self->{temp_schema}[-1], 0);
```
jhthorsen: Extra ", 0" at the end.

lib/JSON/Validator.pm
```diff
@@ -574,69 +573,101 @@ sub _validate {
 
 sub _validate_all_of {
   my ($self, $data, $path, $rules) = @_;
-  my $type = data_type $data, $rules;
-  my (@errors, @expected);
+  my (@errors);
```
jhthorsen: No need for parenthesis.

lib/JSON/Validator.pm
```diff
@@ -574,69 +573,101 @@ sub _validate {
...
+    [
+    allOf => type => join('/', uniq map $_->[1]->details->[0], @errors),
+    $errors[-1][1]->details->[2]
+    ]
+    if @errors > 1 and not grep $_->[1]->details->[1] ne 'type', @errors;
+
+  return prefix_errors allOf => @errors;
+}
+
+sub _validate_any_of_types {
```

lib/JSON/Validator.pm
```diff
@@ -574,69 +573,101 @@ sub _validate {
...
+  my ($self, $data, $path, $rules) = @_;
+  my @errors;
+
+  for my $rule (@$rules) {
+    my @e = $self->_validate($_[1], $path, $rule);
+    return unless @e;
```
jhthorsen: I think you can merge these two lines: `return unless my @e = $self->_validate($_[1], $path, $rule);`, or maybe this even:

```
push @errors, $self->_validate($_[1], $path, $rule);
return unless @errors;
```

lib/JSON/Validator.pm
```diff
@@ -574,69 +573,101 @@ sub _validate {
...
+    push @errors, @e;
+  }
+
+  if (my @e = grep $_->details->[1] ne 'type', @errors) {
+    return @e;
```
jhthorsen: I think this is not what you want. Why filter out the "type" errors? See my comment below for a fixed version.

lib/JSON/Validator.pm
```diff
@@ -574,69 +573,101 @@ sub _validate {
...
+  }
+
+  # combine all 'type' errors together
+  my $want_types = join('/', uniq map $_->details->[0], @errors);
+  return E $path, [$want_types => 'type', $errors[-1]->details->[2]];
```
jhthorsen: I would like to use `@details` in the code above:

```
  my @details = grep { $_->[1] eq 'type' } map { $_->details } @errors;
  return @errors unless @details == @errors;

  my $want_types = join '/', map { $_->[0] } @details;
  return E $path, [$want_types => type => $details[-1][2]];
```

t/jv-allof.t
karenetheridge: I have not changed the sorting in this PR. Let's look at that later.

t/jv-anyof.t
karenetheridge: fixed, by splitting apart the anyOf logic into two subs, and the array-type-specific version will disregard mismatched type errors if there are more specific errors we can report instead.

lib/JSON/Validator.pm
karenetheridge: let's look at combining them afterwards, to keep the changes here simpler. I don't think it will help any (the subs are all pretty short now, but subtly different in how they manage the loop - one of them short-circuits and the others don't; one keeps track of passing tests and the others don't..), so combining them might make readability worse.

lib/JSON/Validator.pm
karenetheridge: I removed most of them but missed this one.

lib/JSON/Validator.pm
karenetheridge: agreed.

lib/JSON/Validator.pm
karenetheridge: Filtering out the type errors resolves the complaint you had about the 'enum' tests, where we were reporting a mismatched type even though the other choice of type did match. (t/jv-anyof.t line 70)

lib/JSON/Validator.pm
karenetheridge: unfortunately that breaks some tests.  I don't see a good way of shortening the code without changing the logic.

jhthorsen:
Thanks for all the changes 👍

PR #202, opened 2020-02-20
Optimize checksum generation for simple cases

EreMaijala:
### Summary
Calling Data::Dumper is relatively slow, so avoiding it when easily possible can provide a significant performance improvement.

All the tests pass, and I've also run them several times with code that chooses the new path randomly only 50% of time. I've also verified that all the cases encountered in the tests and validation of the real schema (see below), the new code and Data::Dumper create equals keys.

### Motivation
Avoiding Data::Dumper for undef, string and numeric variables makes the validation about 40 percent faster for my test case of loading and validating Koha's REST API schema.

### References
My test script: https://gist.github.com/EreMaijala/5ba3ae6e3739c948f774a8a1246ca879
My test results: https://gist.github.com/EreMaijala/163b135c280be694e615c2be4e3ac2f7
Koha's REST API schema: https://github.com/Koha-Community/Koha/blob/master/api/v1/swagger/swagger.json

### DIFF
--- a/lib/JSON/Validator/Util.pm
+++ b/lib/JSON/Validator/Util.pm
@@ -2,7 +2,6 @@ package JSON::Validator::Util;
 use Mojo::Base -strict;
 
 use Carp         ();
-use Data::Dumper ();
 use Exporter 'import';
 use JSON::Validator::Error;
 use List::Util;
@@ -18,7 +17,11 @@ our @EXPORT_OK
 sub E { JSON::Validator::Error->new(@_) }
 
 sub data_checksum {
-  Mojo::Util::md5_sum(Data::Dumper->new([@_])->Sortkeys(1)->Useqq(1)->Dump);
+  return Mojo::Util::md5_sum(
+    ref $_[0]
+    ? Mojo::JSON::encode_json([@_])
+    : defined $_[0] ? qq('$_[0]')
+    :                 'undef');
 }
 
 sub data_section {
@@ -118,7 +121,7 @@ sub prefix_errors {
 }
 
 sub schema_type {
-  return '' if ref $_[0] ne 'HASH';
+  return ''            if ref $_[0] ne 'HASH';
   return $_[0]->{type} if $_[0]->{type};
   return _guessed_right(object => $_[1]) if $_[0]->{additionalProperties};
   return _guessed_right(object => $_[1]) if $_[0]->{patternProperties};
--- a/lib/JSON/Validator/Util.pm
+++ b/lib/JSON/Validator/Util.pm
@@ -1,7 +1,7 @@
 package JSON::Validator::Util;
 use Mojo::Base -strict;
 
-use Carp         ();
+use Carp ();
 use Exporter 'import';
 use JSON::Validator::Error;
 use List::Util;
@@ -10,18 +10,19 @@ use Mojo::JSON;
 use Mojo::Loader;
 use Mojo::Util;
 use Scalar::Util 'blessed';
+use YAML::XS;
 
 our @EXPORT_OK
   = qw(E data_checksum data_section data_type is_type schema_extract json_pointer prefix_errors schema_type);
 
 sub E { JSON::Validator::Error->new(@_) }
 
+my $serializer
+  = eval 'use Sereal::Encoder;1' ? \&_sereal_encode : \&YAML::XS::Dump;
+
 sub data_checksum {
   return Mojo::Util::md5_sum(
-    ref $_[0]
-    ? Mojo::JSON::encode_json([@_])
-    : defined $_[0] ? qq('$_[0]')
-    :                 'undef');
+    ref $_[0] ? $serializer->($_[0]) : defined $_[0] ? qq('$_[0]') : 'undef');
 }
 
 sub data_section {
@@ -192,6 +193,11 @@ sub _schema_extract {
 
   return $cb->($data, $pos) if $cb;
   return $data;
+}
+
+sub _sereal_encode {
+  state $s = Sereal::Encoder->new({canonical => 1});
+  return $s->encode($_[0]);
 }
 
 1;
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -31,7 +31,7 @@ my %WriteMakefileArgs = (
     },
   },
   PREREQ_PM     => {'List::Util' => '1.45', 'Mojolicious' => '7.28'},
-  TEST_REQUIRES => {'Test::More' => '1.30', 'Test::Deep'  => '0'},
+  TEST_REQUIRES => {'Test::More' => '1.30', 'Test::Deep'  => '0', 'Test::Without::Module' => '0.20'},
   test => {TESTS => (-e 'META.yml' ? 't/*.t' : 't/*.t xt/*.t')},
 );
 
--- a/t/util-checksum-yaml-xs.t
+++ b/t/util-checksum-yaml-xs.t
@@ -1,0 +1,29 @@
+use Test::Without::Module qw( Sereal::Encoder );
+
+use Mojo::Util 'md5_sum';
+use JSON::Validator;
+use JSON::Validator::Util qw(data_checksum);
+use Test::More;
+
+my $d_hash  = {foo => {}, bar => {}};
+my $d_hash2 = {bar => {}, foo => {}};
+my $d_undef = {foo => undef};
+my $d_obj   = {foo => JSON::Validator::Error->new};
+my $d_array  = ['foo', 'bar'];
+my $d_array2 = ['bar', 'foo'];
+
+use_ok 'YAML::XS';
+isnt data_checksum($d_array), data_checksum($d_array2), 'data_checksum array';
+is data_checksum($d_hash), data_checksum($d_hash2),
+  'data_checksum hash field order';
+isnt data_checksum($d_hash), data_checksum($d_undef),
+  'data_checksum hash not undef';
+isnt data_checksum($d_hash), data_checksum($d_obj),
+  'data_checksum hash not object';
+isnt data_checksum($d_obj), data_checksum($d_undef),
+  'data_checksum object not undef';
+isnt data_checksum(3.14), md5_sum(3.15), 'data_checksum numeric';
+is data_checksum(3.14), data_checksum('3.14'),
+  'data_checksum numeric like string';
+
+done_testing;
--- a/t/util.t
+++ b/t/util.t
@@ -1,8 +1,9 @@
 use Mojo::Base -strict;
 use Mojo::JSON 'false';
+use Mojo::Util 'md5_sum';
 use JSON::Validator;
 use JSON::Validator::Util
-  qw(E data_type schema_type prefix_errors is_type json_pointer);
+  qw(E data_checksum data_type schema_type prefix_errors is_type json_pointer);
 use Test::More;
 
 my $e = E '/path/x', 'some error';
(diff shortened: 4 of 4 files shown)


lib/JSON/Validator/Util.pm
```diff
@@ -18,7 +18,22 @@ our @EXPORT_OK
...
+      $key = '$VAR1 = ' . $_[0] . ';';
+    }
+    else {
+      $key = '$VAR1 = "' . $_[0] . '";';
+    }
+  }
+  else {
+    $key = Data::Dumper->new([@_])->Sortkeys(1)->Useqq(1)->Indent(0)->Dump;
+  }
+  Mojo::Util::md5_sum($key);
```
jhthorsen: I think this can be condensed quite a bit:

```
sub data_checksum {
  return Mojo::Util::md5_sum(
    ref $_[0]
    ? Data::Dumper->new([@_])->Indent(0)->Sortkeys(1)->Terse(1)->Useqq(1)->Dump
    : defined $_[0] ? qq("$_[0]")
    :                 'undef');
}
```

EreMaijala: Works for me. Want me to do it or would you rather do that yourself since there's not much of my code anymore :) ?

EreMaijala:
Begs the question why that is, but I couldn't spend more time digging any deeper. I profiled that Dump calls Dumpxs which should be fast native code, but it's obviously not very fast for these cases.

lib/JSON/Validator/Util.pm
jhthorsen: It is your code though, just formatted a bit differently 😄It all depends on if you want your name in the code or not I guess - You decide 👍

jhthorsen:
I was wondering if we could use Sereal, Storable or something else, but I haven't checked if they will make the same data structure / checksum on each run.

lib/JSON/Validator/Util.pm
EreMaijala: Ok, updated. :) I also had to reintroduce the numeric check, otherwise tests would fail.

EreMaijala:
I'll check Sereal and Storable out quickly...

lib/JSON/Validator/Util.pm
jhthorsen: Oh. That's terrible. In that case, I think you need to use Scalar::Util::looks_like_number(), because your regex fails on `3.14`.

EreMaijala: I deliberately made it work like Data::Dumper, which would also return 3.14 as `'3.14'`.

EreMaijala:
Storable won't do it since it can't sort hash keys. Sereal would work and seems to be about as fast for my test case. Want me to switch?

jhthorsen:
I forgot about YAML::XS. I'm going to depend on that in future versions anyways, so maybe that's better to use? It's fine if it's not _as_ fast as Sereal. It just can't be crazy slow like Dumper.

EreMaijala:
YAML::XS is usable but a bit slower, in my test case it takes ~6.8 seconds vs ~6.1 seconds. I could live with that, but it would be hard as I tested also the custom code + YAML::XS instead of Data::Dumper, and it would give ~5.5 seconds..

lib/JSON/Validator/Util.pm
jhthorsen: Can you tell me which test that fail without the numeric test?

jhthorsen:
I think I want to go with `yaml_shortcut()` in the example below. Either way, this change needs some more tests to get merged. Once the tests are in place, then we can also opt in for Sereal.

The tests cannot check the returned MD5 sum, since it will change when we change from Data::Dumper to YAML::XS or Sereal. Instead the test need to check that objects are not ignored, and the number logic works. I'm not sure if the dummy test below works, but take it as a rough example:

```
my $d_hash  = {foo => {}};
my $d_undef = {foo => undef};
my $d_obj   = {foo => JSON::Validator::Error->new};
isnt data_checksum($d_hash), data_checksum($d_undef);
isnt data_checksum($d_hash), data_checksum($d_obj);
isnt data_checksum($d_obj), data_checksum($d_undef);
isnt data_checksum(3.14), md5_sum(3.15);
```

One way to check if the tests do what they are supposed to could be to replace Data::Dumper with Mojo::JSON. Mojo::JSON will ignore the JSON::Validator::Error object, which should make one of the `isnt` tests fail.

Btw, below are my benchmark results. Sereal is so incredible fast, but I think in many cases yaml_shortcut() will be fast enough. (I don't want to have too many dependencies, if I can avoid it)

```
use Mojo::Base -strict;
use Data::Dumper ();
use Mojo::Util 'md5_sum';
use Sereal::Encoder ();
use YAML::XS        ();

sub dd {
  return md5_sum(
    Data::Dumper->new([$_[0]])->Indent(0)->Sortkeys(1)->Terse(1)->Useqq(1)
      ->Dump);
}

my $s = Sereal::Encoder->new({canonical => 1, sort_keys => 1});
sub sereal { return md5_sum($s->encode($_[0])) }

sub yaml { return md5_sum(YAML::XS::Dump($_[0])) }

sub dd_shortcut {
  return md5_sum 'undef' unless defined $_[0];
  return md5_sum $_[0] =~ /^\d+$/ ? $_[0] : qq("$_[0]") unless ref $_[0];
  return md5_sum(
    Data::Dumper->new([$_[0]])->Indent(0)->Sortkeys(1)->Terse(1)->Useqq(1)
      ->Dump);
}

sub yaml_shortcut {
  return md5_sum 'undef' unless defined $_[0];
  return md5_sum $_[0] =~ /^\d+$/ ? $_[0] : qq("$_[0]") unless ref $_[0];
  return md5_sum(YAML::XS::Dump($_[0]));
}

bm(hash   => {foo => 42});
bm(undef  => undef);
bm(number => 42);
bm(string => 'foo');

sub bm {
  my ($desc, $data) = @_;
  use Benchmark 'cmpthese';

  my %tests;
  for my $name (qw(dd dd_shortcut yaml yaml_shortcut sereal)) {
    my $code = main->can($name) or die $name;
    $tests{$name} = sub { $code->($data) };
  }

  warn "\n#--- $desc\n";
  cmpthese($ENV{COUNT} || 500_000, \%tests);
}

__DATA__
$ COUNT=500000 perl b.pl
#--- hash
                  Rate dd_shortcut         dd yaml_shortcut       yaml    sereal
dd_shortcut    77399/s          --        -2%          -53%       -54%      -90%
dd             78740/s          2%         --          -53%       -54%      -90%
yaml_shortcut 166113/s        115%       111%            --        -2%      -79%
yaml          170068/s        120%       116%            2%         --      -78%
sereal        781250/s        909%       892%          370%       359%        --

#--- undef
            (warning: too few iterations for a reliable count)
            (warning: too few iterations for a reliable count)
                   Rate        dd       yaml    sereal yaml_shortcut dd_shortcut
dd             105263/s        --       -61%      -91%          -96%        -96%
yaml           270270/s      157%         --      -78%          -89%        -89%
sereal        1219512/s     1059%       351%        --          -51%        -51%
yaml_shortcut 2500000/s     2275%       825%      105%            --          0%
dd_shortcut   2500000/s     2275%       825%      105%            0%          --

#--- number
            (warning: too few iterations for a reliable count)
            (warning: too few iterations for a reliable count)
                   Rate        dd       yaml    sereal yaml_shortcut dd_shortcut
dd             101420/s        --       -59%      -92%          -93%        -94%
yaml           245098/s      142%         --      -80%          -82%        -85%
sereal        1219512/s     1102%       398%        --          -12%        -24%
yaml_shortcut 1388889/s     1269%       467%       14%            --        -14%
dd_shortcut   1612903/s     1490%       558%       32%           16%          --

#--- string
            (warning: too few iterations for a reliable count)
                   Rate        dd       yaml    sereal dd_shortcut yaml_shortcut
dd             100200/s        --       -57%      -92%        -92%          -93%
yaml           230415/s      130%         --      -81%        -82%          -83%
sereal        1219512/s     1117%       429%        --         -2%          -12%
dd_shortcut   1250000/s     1147%       442%        2%          --          -10%
yaml_shortcut 1388889/s     1286%       503%       14%         11%            --
```

lib/JSON/Validator/Util.pm
EreMaijala: Uh, err... none. I could swear there was something but I must have made a mistake somewhere. Perhaps I had the "randomly select Data::Dumper instead" debug code there at that time.

EreMaijala:
Nice benchmark! Sereal is awesomely fast, but I understand the desire to minimize dependencies. Latest commit switches to YAML::XS and adds tests (yours and a couple more). I verified that the tests will fail if MOJO::JSON or Storable is used. Does something else need to be changed since YAML::XS becomes a hard dependency?

karenetheridge:
> Mojo::JSON will ignore the JSON::Validator::Error object, which should make one of the isnt tests fail

But we're never doing checksums on Error objects, are we? So this shouldn't be important.  I was wondering how YAML::XS compared to Mojo::JSON (which is really Cpanel::JSON::XS beneath) for speed.

EreMaijala:
> But we're never doing checksums on Error objects, are we? So this shouldn't be important. I was wondering how YAML::XS compared to Mojo::JSON (which is really Cpanel::JSON::XS beneath) for speed.

I can't say I know much about that, but several tests failed with Mojo::JSON. Perhaps it's the boolean class?

karenetheridge:
That would depend on how the tests failed? :)

EreMaijala:
@karenetheridge Well, err, uh.. I must have made a mistake somewhere when I tested Mojo:JSON. It's actually working fine, is as fast as YAML::XS and all the tests are passing. The latest patch now uses it, so no hard dependency on YAML::XS needed for this. Thanks for your perseverance on this! :)

lib/JSON/Validator/Util.pm
```diff
@@ -18,7 +17,11 @@ our @EXPORT_OK
 sub E { JSON::Validator::Error->new(@_) }
 
 sub data_checksum {
-  Mojo::Util::md5_sum(Data::Dumper->new([@_])->Sortkeys(1)->Useqq(1)->Dump);
+  return Mojo::Util::md5_sum(
+    ref $_[0]
+    ? Mojo::JSON::encode_json([@_])
```
jhthorsen: I don't trust this. I used Data::Dumper for a reason, I cannot remember which case failed, but something has failed before. It's a bit annoying, since it seems like that code was in Swagger2::SchemaValidator, which is history 😥

lib/JSON/Validator/Util.pm
```diff
@@ -11,14 +10,19 @@ use Mojo::JSON;
...
 
 sub E { JSON::Validator::Error->new(@_) }
 
 sub data_checksum {
-  Mojo::Util::md5_sum(Data::Dumper->new([@_])->Sortkeys(1)->Useqq(1)->Dump);
+  return Mojo::Util::md5_sum(
+    ref $_[0]
+    ? Dump([@_])
+    : defined $_[0] ? qq('$_[0]')
+    :                 'undef');
```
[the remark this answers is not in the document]
EreMaijala: ok

lib/JSON/Validator/Util.pm
EreMaijala: Ok. Now using Sereal::Encoder if available and YAML::XS if not.

EreMaijala:
I resorted to Test::Without::Module as I couldn't make the @INC manipulation to work (no matter what I tried, the sub didn't get called).

jhthorsen:
I’ll fix the @INC part. Thanks for work! This is great 👍

jhthorsen:
I’ll fix the @INC part. Thanks for work! This is great 👍

PR #203, opened 2020-02-21
fix edge cases from PR#201 where error types are misidentified

karenetheridge:
### Summary
These new test cases expose cases where the `[ $i, @errors ]` structures have more than one error in them, and consequently the detection of 'type' errors was incorrect.  A change to JSON::Validator::Error was required to preserve the `details` field that we rely on here (`prefix_errors` was removing that field, resulting in the errors appearing to have `generic` type.)

### References
as discussed on irc from the last few days.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -586,13 +586,15 @@ sub _validate_all_of {
 
   return if not @errors;
 
+  my @details = map { my ($i, @e) = @$_; map $_->details, @e } @errors;
+
   return prefix_errors(allOf => @errors)
-    if @errors == 1
-    or grep $_->[1]->details->[1] ne 'type', @errors;
+    if @details == 1
+    or grep $_->[1] ne 'type', @details;
 
   # combine all 'type' errors together
-  my $want_types = join('/', uniq map $_->[1]->details->[0], @errors);
-  return E $path, [allOf => type => $want_types, $errors[-1][1]->details->[2]];
+  my $want_types = join '/', uniq map $_->[0], @details;
+  return E $path, [allOf => type => $want_types, $details[-1][2]];
 }
 
 sub _validate_any_of_types {
@@ -610,8 +612,9 @@ sub _validate_any_of_types {
   }
 
   # combine all 'type' errors together
-  my $want_types = join('/', uniq map $_->details->[0], @errors);
-  return E $path, [$want_types => 'type', $errors[-1]->details->[2]];
+  my @details    = map $_->details, @errors;
+  my $want_types = join '/', uniq map $_->[0], @details;
+  return E $path, [$want_types => 'type', $details[-1][2]];
 }
 
 sub _validate_any_of {
@@ -627,12 +630,15 @@ sub _validate_any_of {
     $i++;
   }
 
+  my @details = map { my ($i, @e) = @$_; map $_->details, @e } @errors;
+
   return prefix_errors(anyOf => @errors)
-    if grep $_->[1]->details->[1] ne 'type', @errors;
+    if @details == 1
+    or grep $_->[1] ne 'type', @details;
 
   # combine all 'type' errors together
-  my $want_types = join('/', uniq map $_->[1]->details->[0], @errors);
-  return E $path, [anyOf => type => $want_types, $errors[-1][1]->details->[2]];
+  my $want_types = join '/', uniq map $_->[0], @details;
+  return E $path, [anyOf => type => $want_types, $details[-1][2]];
 }
 
 sub _validate_one_of {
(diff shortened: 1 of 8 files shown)


lib/JSON/Validator.pm
```diff
@@ -586,13 +586,15 @@ sub _validate_all_of {
 
   return if not @errors;
 
+  my @details = map { map $_->details, @{$_}[1 .. $#{$_}] } @errors;
```
jhthorsen: This is too hard to read. I don't like nested maps, and there's just too much {@[$#]}. (Same below)

lib/JSON/Validator.pm
```diff
@@ -586,13 +586,15 @@ sub _validate_all_of {
...
-    if @errors == 1
-    or grep $_->[1]->details->[1] ne 'type', @errors;
+    if @details == 1
+    or grep $_->[1] ne 'type', @details;
 
   # combine all 'type' errors together
-  my $want_types = join('/', uniq map $_->[1]->details->[0], @errors);
-  return E $path, [allOf => type => $want_types, $errors[-1][1]->details->[2]];
+  my $want_types = join('/', uniq map $_->[0], @details);
+  return E $path, [allOf => type => $want_types, $details[-1]->[2]];
```
jhthorsen: No need for the last `->`, and `join` does not need any parenthesis. (Same below)

lib/JSON/Validator.pm
```diff
@@ -586,13 +586,15 @@ sub _validate_all_of {
 
   return if not @errors;
 
+  my @details = map { map $_->details, @{$_}[1 .. $#{$_}] } @errors;
+
   return prefix_errors(allOf => @errors)
-    if @errors == 1
-    or grep $_->[1]->details->[1] ne 'type', @errors;
+    if @details == 1
+    or grep $_->[1] ne 'type', @details;
```
jhthorsen: I also find grep without {} hard to read, especially when it's not just a simple expression.

`$_->[1]` == simple
`$_->[1] ne 'type'` != simple

lib/JSON/Validator/Util.pm
```diff
@@ -110,7 +110,7 @@ sub prefix_errors {
     push @errors, map {
       my $msg = sprintf '/%s/%s %s', $type, $index, $_->message;
       $msg =~ s!(\d+)\s/!$1/!g;
-      E $_->path, $msg;
+      E +{%$_, message => $msg};   # preserve 'details', for later introspection
```
jhthorsen: 👍

t/jv-allof.t
```diff
@@ -42,7 +42,18 @@ validate_ok 'he110th3re', $schema,
...
-  E('/', '/allOf/1 Expected integer/boolean - got string.');
+validate_ok 'hello', {allOf => [{type => ['integer', 'boolean']}]},
+  E('/', '/allOf/0 Expected integer/boolean - got string.');
+
+validate_ok 'hello',
+  {
+  allOf => [
+    {allOf => [{type => 'boolean'}, {type => 'string', maxLength => 2}]},
+    {type  => 'integer'},
+  ],
```
jhthorsen: Something is up with perltidy here 😞

lib/JSON/Validator.pm
karenetheridge: I'm not sure what I can do with that, as the code is correct now :)  I can add more local variables, which would make the expression longer.. I'm not sure if it would be any easier to read.

lib/JSON/Validator.pm
karenetheridge: fixed.

t/jv-allof.t
karenetheridge: I thought so, but isn't this how you like it? ;/   is the .perltidyrc incorrect?

karenetheridge: I am definitely using the local .perltidyrc with the latest Perl::Tidy (version 20200110).

lib/JSON/Validator.pm
karenetheridge: I'll push a change with some local vars showing how the list is being chopped up and see if that makes it better.

karenetheridge:
I thought of more test cases that exposed more edge cases ( 😃  and 😲 ) which I added in a separate commit.  It may be better to squash these together.  Also, it may be preferred now to invert the (or..not or not..) conditionals in the `_validate_*_of` subs, so we do the collapse-types-together path first for readability. I kept it the way it is now because the diffs are smaller, and so you can see what I changed, but I'm happy to make that change too and squash it all together.

karenetheridge:
For the second commit, this definitely fixes a bug, but I wonder if it exposes another one...

this test case:

```
validate_ok {foo => 'not an arrayref'},
  {
  anyOf => [
    {type => 'object', properties => {foo => {type => 'array'}}},
    {type => 'boolean'},
  ]
  },
  E('/foo', '/anyOf/0 Expected array - got string.'),
  E('/',    '/anyOf/1 Expected boolean - got object.');
```

without the fix, we got a totally nonsensical error message..  but I wonder if we should only return the first error message (the one that doesn't pertain to an incorrect type at the top level of the data), so behaviour would be the same as if we had said `type => ['object','boolean']`.

To fix that, I can just discard the errors where the error type is 'type' and the path eq $path, which is already the check being done to see how to fold errors together.

t/jv-allof.t
jhthorsen: Sorry, nothing against your code.

jhthorsen:
I had to merge this manually, since there was some commits that does not seem to be rebased correctly.

PR #207, opened 2020-03-16
fixes for arrays

karenetheridge:
### Summary
Fixes for validating the 'array' type.

### Motivation
The new unit tests do not pass without the accompanying code changes.

### References
https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.4

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -786,6 +786,7 @@ sub _validate_type_array {
       push @e, \@tmp if @tmp;
     }
     push @errors, map {@$_} @e if @e >= @$data;
+    push @errors, E $path, [array => 'contains'] if not @$data;
   }
   elsif (ref $schema->{items} eq 'ARRAY') {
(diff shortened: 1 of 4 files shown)


lib/JSON/Validator.pm
```diff
@@ -753,8 +753,6 @@ sub _validate_format {
   return E $path, [format => $schema->{format}, $err];
 }
 
-sub _validate_type_any { }
```
jhthorsen: I don't think this change is relevant for this PR? This also goes for the change in jv-basic.t.

karenetheridge: sure, I can pull it out and move it to a later PR if you like. The only use of the 'any' type occurs within array validation, so I thought it was reasonable to include it here.

karenetheridge: and done.

jhthorsen:
Thanks!

PR #208, opened 2020-04-01
Fix uri resolution

karenetheridge:
### Summary
There are various cases where uri resolution was not performed correctly with respect to `$id` and `$ref` keywords.
Many test cases have been added demonstrating things that did not work previously.

### Motivation
My application that uses JSON::Validator needs to pre-load and use documents with `$ref`s like `/foo/bar/baz#/json/pointer` and resolve them against `$id`s like `http://spec.app.corp.com`.

### References
URI resolution within document traversal is specified at https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.

Thank you Joyent for sponsoring the approximately 3 days of paid work that went into producing and verifying this pull request.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -394,16 +394,18 @@ sub _register_schema {
 sub _resolve {
   my ($self, $schema) = @_;
   my $id_key = $self->_id_key;
-  my ($id, $resolved);
+  my $id;
 
   local $self->{level} = $self->{level} || 0;
   delete $_[0]->{schemas}{''} unless $self->{level};
 
   if (ref $schema eq 'HASH') {
     $id = $schema->{$id_key} // '';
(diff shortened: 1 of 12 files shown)


karenetheridge:
(the force-push I just did was only correcting a typo in the last commit message.)

jhthorsen:
Could you implement this functionality on top of #211? I think #211 is more aligned with what you need, even though it doesn't fix the issues you have here.

...maybe I should merge #210 before asking to build on top of #211 ;)

karenetheridge:
No, #210 and #211 change a lot more things, and will undo the fixes that I am making here (and make it a *lot* harder to fix them again on top). If you merge #208 first you will see the issues with your refactoring.

jhthorsen:
I think this PR is now outdated after adding JSON::Validator::Store.

PR #189, opened 2020-02-11
Add OpenAPI schema classes

jhthorsen:
### Summary
This change aims to make it possible to move logic from JSON::Validator into the different JSON::Validator::Schema subclasses, but at the same time be backward compatible. After merging this PR, I need to update Mojolicious::Plugin::OpenAPI before moving the actual validation methods. (I think)

### Motivation
Same motivation as #209.

### References
#209 and #210

PR #209, opened 2020-04-10
Change JSON::Validator::schema() to hold a JSON::Validator::Schema object

jhthorsen:
### Summary
This PR is to make #189 smaller and easier to read. This PR is also back compatible with other modules such as JSON::Validator::OpenAPI::Mojolicious.

### Motivation
I think it's necessarily to split the functionality of JSON::Validator into separate schema classes to reduce the cost of maintaining the code for the new JSON-Schema drafts (and OpenAPI specifications. This PR is just the first step in that direction. Right now, we are getting more and more if/else branches such as `if ($jv->version == 6)` to run the correct logic for a given schema. I think creating a new JSON::Validator::Schema sub class will make it easier to read and understand the code.

The next step, without considering #189, would be to allow all the schema objects to be part of a shared pool, so if J::V creates a couple of schema objects, then they should be able to reference each other. I think this is pretty much done internally with `$self->{schemas}`, but it needs to be exposed in a nice way to the user.

### References
#189

### DIFF
--- a/Changes
+++ b/Changes
@@ -7,6 +7,8 @@ Revision history for perl distribution JSON-Validator
  - Deprecated JSON::Validator::version()
  - Removed JSON::Validator::generate_definitions_path()
  - Removed support for JSON::Validator::bundle({ref_key => ...})
+ - JSON::Validator::schema() now holds a JSON::Validator::Schema object
+   instead of Mojo::JSON::Pointer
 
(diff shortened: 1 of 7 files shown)


jhthorsen:
I'm quite sure this is a good move, after all the discussion on irc://irc.freenode.net/%23perl-openapi and the comments in #189.

PR #211, opened 2020-04-16
[WIP] Add JSON::Validator::Store

jhthorsen:
### Summary
Adding a store that holds the raw format of the schemas. Note that this PR is work-in-progress, and I mostly opened it to get feedback on the idea.

### Motivation
Having a schema store will allow users to add schemas with "$id", that might be referenced later on when resolving a schema with "$ref" that points to a given "$id". Having a store class, will also allow users to subclass the store to load schemas from a database, like this:

```perl
package JSON::Validator::Store::DBI;
use Mojo::Base 'JSON::Validator::Store';

sub load_schema {
  my ($self, $url) = @_;
  my $sth = $self->dbh->prepare("select schema from schemas where url = ?", $url);
  $sth->execute($url);
  my $row = $sth->fetchrow_arrayref;
  return $self->load_schema_from_text(\$row->[0]) if $row;
  return $self->SUPER::load_schema($url);
}
```

Having a custom store can also disallow loading schemas from disk and/or web.

### Questions
1. Should the schemas be (unresolved) JSON::Validator::Schema objects?
2. Should probably add load_schema_p() (for non-blocking support)


jhthorsen:
Going to close this PR for now. Might revisit the idea later on...

PR #212, opened 2020-04-23
demonstration: broken schema registration

karenetheridge:
This demonstrates the problem with broken schema caching and resolution.

I fixed the simple problem of the registry being lost when ::Schema objects are changed (see its commit message), but because of the order in which the new Schema object is constructed, ref resolution ends in an infinite loop.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -370,6 +370,7 @@ sub _new_schema {
   my $schema = $self->_schema_class($spec)->new(
     $spec, @attrs,
     version => $self->{version},
+    schemas => ($self->{schemas} //= {}),
     map { ($_ => $self->$_) } qw(cache_paths formats ua)
   );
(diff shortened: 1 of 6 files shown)


jhthorsen:
As mentioned in other channels: I think you should convert/merge whatever I started on in #211 with the code in #208.

I'm still keeping this PR open though, so I/we don't forget about it.

jhthorsen:
Thanks for pointing me in the right direction. This is now fixed in "master".

karenetheridge:
Yes, I provided you a fix for that months ago in another PR :)

jhthorsen:
> Yes, I provided you a fix for that months ago in another PR :)

Where? And why was this PR still open then?

karenetheridge:
If I recall, you wanted me to reimplement it on top of some rearchitecting you were doing in a side branch, but doing so would have made the fix even harder.

I guess it doesn't really matter anymore.

PR #214, opened 2020-05-08
Fix the coercion of boolean value 'false'

SebMourlhou:
With "coerce bool" activated, the result of the coercion of the string 'false' should be false and not true.


jhthorsen:
Oh! This is really bad 😞 

I've fixed this in c64c8fb, since I also wanted to coerce the strings "0" and "1".

PR #215, opened 2020-05-09
Allow booleans to be scalar refs

waterkip:
The JSON validation breaks when one uses `\1` or `\0` when using the
serializion in conjuction with JSON::XS, Mojo::JSON and others.

Signed-off-by: Wesley Schwengle <wesley@opndev.io>

### DIFF
--- a/lib/JSON/Validator/Util.pm
+++ b/lib/JSON/Validator/Util.pm
@@ -57,13 +57,23 @@ sub data_section {
   Carp::confess(qq(Could not find "$file" in __DATA__ section of $err.));
 }
 
+sub _is_boolean {
+  my $ref     = ref $_[0];
+  my $blessed = blessed $_[0];
+  return 1 if $blessed && ("$_[0]" eq "1" or !"$_[0]");
+  return 1 if $blessed && $_[0]->isa('JSON::PP::Boolean');
+  return 1 if $ref eq 'SCALAR' && (${$_[0]} == 1 || ${$_[0]} == 0);
+  return 0;
+}
+
+
 sub data_type {
   my $ref     = ref $_[0];
   my $blessed = blessed $_[0];
   return 'object'  if $ref eq 'HASH';
-  return lc $ref   if $ref and !$blessed;
   return 'null'    if !defined $_[0];
-  return 'boolean' if $blessed and ("$_[0]" eq "1" or !"$_[0]");
+  return 'boolean' if _is_boolean($_[0]);
+  return lc $ref   if $ref and !$blessed;
 
   if (is_type($_[0], 'NUM')) {
     return 'integer' if grep { ($_->{type} // '') eq 'integer' } @{$_[1] || []};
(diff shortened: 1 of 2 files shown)


jhthorsen:
I don't see any reason to accept this, since scalar refs is not a thing in JSON. Please provide a better argument for why/when this is needed.

waterkip:
```
$ cat json-bool.pl; perl json-bool.pl
#!/usr/bin/perl
use warnings;
use strict;
require JSON::XS;
require JSON::PP;
require Mojo::JSON;
require Cpanel::JSON::XS;

print JSON::XS::encode_json({ foo => \1 }), $/;
print JSON::PP::encode_json({ foo => \1 }), $/;
print Mojo::JSON::encode_json({ foo => \1 }), $/;
print Cpanel::JSON::XS::encode_json({ foo => \1 }), $/;
print JSON::XS::encode_json({ foo => \0 }), $/;
print JSON::PP::encode_json({ foo => \0 }), $/;
print Mojo::JSON::encode_json({ foo => \0 }), $/;
print Cpanel::JSON::XS::encode_json({ foo => \0 }), $/;

{"foo":true}
{"foo":true}
{"foo":true}
{"foo":true}
{"foo":false}
{"foo":false}
{"foo":false}
{"foo":false}
```

They all serializer \1 and \0 to true and false. Which is why I think the validation must respect this syntax as well.

jhthorsen:
You still haven't given me any argument for when this is beneficial.

waterkip:
If I use \1 or \0 in code somewhere and I pass it to the Mojolious Plugin for OpenAPI and I set it to validate the JSON before sending it out it croaks. I dug around and saw that the JSON validator which is used by that plugin doesn't allow this syntax. So when I try to validate the output the output is considered invalid while when I disable the validation it works fine and returns a boolean value as shown in the json-bool.pl example with the various serializers.

Which leads me to 1) Not use the validation 2) Patch the validator that is accepts such values. I opted for 2 because I want to know when I do something wrong. Another workaround would be to serialize to json, deserialize it and then pass it to the validator after which another round of serialization happens. Which I find a bit counter intuitive.

jhthorsen:
Ooops! Closed the wrong issue :(

jhthorsen:
You should really just use `false` and `true` in that case. There's no way to send a scalar ref using JSON, so I'm pretty sure this is the wrong fix.

Keeping it open a bit longer though in case other people have relevant information.

jhthorsen:
And when I say `false` and `true`, I mean functions such as `YAML::XS::false` or `Mojo::JSON::true` (and friends)

waterkip:
From the documentation of Mojo::JSON:
> In addition scalar references will be used to generate booleans, based on if their values are true or false.

From the documentation of Types::Serializer (used by JSON::XS):
> The recommended way to detect whether a scalar is one of these objects is to check whether the stash is the Types::Serialiser::Boolean or Types::Serialiser::Error stash, and then follow the scalar reference to see if it's 1 (true), 0 (false) or undef (error).

I think Mojo::JSON is used by Mojo and thus by the OpenAPI plugin. So you are saying you don't want to support that native serializer? Seems odd but ok.

karenetheridge:
Mojo::JSON doesn't use JSON::XS.

waterkip:
I never said that. I just quoted their documentation and the docs of JSON::XS.

jhthorsen:
Going to close this issue, since I don't agree with the arguments given.

PR #220, opened 2020-08-04
Adds an option to disable the check that tracks the data structure being validated for recursion

jlcooper:
### Summary
This pull request adds an attribute that can be used to deactivate the active recursive data protection logic for a JSON::Validator instance.

### Motivation
When validating a data structure the recursive data protection has to track the references seen as a way to avoid infinite loops. When validating a large data structure tracking the references can use a significant amount of memory and if the caller is willing to guarantee that their data structure doesn't contain any loops then being able to disable it can reduce their process's memory usage.

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -342,7 +342,7 @@ sub _load_schema_from_url {
 
   my $tx  = $self->ua->get($url);
   my $err = $tx->error && $tx->error->{message};
-  confess "GET $url == $err"               if DEBUG and $err;
+  confess "GET $url == $err" if DEBUG and $err;
   die "[JSON::Validator] GET $url == $err" if $err;
 
   if ($cache_path
@@ -518,11 +518,11 @@ sub _resolve_ref {
     last if !$ref or ref $ref;
     $fqn = $ref =~ m!^/! ? "#$ref" : $ref;
     my ($location, $pointer) = split /#/, $fqn, 2;
-    $url     = $location = _location_to_abs($location, $url);
+    $url = $location = _location_to_abs($location, $url);
     $pointer = undef if length $location and !length $pointer;
     $pointer = url_unescape $pointer if defined $pointer;
-    $fqn     = join '#', grep defined, $location, $pointer;
-    $other   = $self->_resolve($location);
+    $fqn   = join '#', grep defined, $location, $pointer;
+    $other = $self->_resolve($location);
 
     if (defined $pointer and length $pointer and $pointer =~ m!^/!) {
       $other = Mojo::JSON::Pointer->new($other)->get($pointer);
@@ -571,7 +571,7 @@ sub _validate {
 
   my @errors;
 
-  if ( $self->recursive_data_protection ) {
+  if ($self->recursive_data_protection) {
     my $seen_addr = join ':', refaddr($schema),
       (ref $data ? refaddr $data : ++$self->{seen}{scalar});
 
@@ -904,7 +904,7 @@ sub _validate_type_integer {
   my @errors = $self->_validate_type_number($_[1], $path, $schema, 'integer');
 
   return @errors if @errors;
-  return         if $value =~ /^-?\d+$/;
+  return if $value =~ /^-?\d+$/;
   return E $path, [integer => type => data_type $value];
 }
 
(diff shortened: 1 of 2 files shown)


lib/JSON/Validator.pm
```diff
@@ -42,6 +42,8 @@ has cache_paths => sub {
 
 has formats => sub { shift->_build_formats };
 
+has recursive_data_protection => sub { return 1 };
```
jhthorsen, line 45: No need for sub, nor return. Just do `has recursive_data_protection => 1;`

lib/JSON/Validator.pm
```diff
@@ -567,15 +569,20 @@ sub _validate {
...
+  my @errors;
+
+  if ( $self->recursive_data_protection ) {
+    my $seen_addr = join ':', refaddr($schema),
+      (ref $data ? refaddr $data : ++$self->{seen}{scalar});
+
+    # Avoid recursion
+    if ($self->{seen}{$seen_addr}) {
+      return @{$self->{seen}{$seen_addr}};
+    }
```
jhthorsen, line 581: Please put this block on one line to save LOC.

lib/JSON/Validator.pm
```diff
@@ -1260,6 +1267,23 @@ block should return C<undef> on success and an error string on error:
 
 See L<JSON::Validator::Formats> for a list of supported formats.
 
+=head2 recursive_data_protection
+
+  my $jv = $jv->recursive_data_protections( $scalar );
+  my $current_state = $jv->recursive_data_protection;
```
jhthorsen, line 1273: Use $boolean instead of $scalar/$current_state to indicate that this is a true/false value.

lib/JSON/Validator.pm
```diff
@@ -1260,6 +1267,23 @@ block should return C<undef> on success and an error string on error:
...
+
+Recursive data protection is active by default, however it can be deactivated
+by assigning a false value to the C<recursive_data_protection> attribute.
```
jhthorsen, line 1276: Use `L</recursive_data_protection>` instead.

lib/JSON/Validator.pm
```diff
@@ -1260,6 +1267,23 @@ block should return C<undef> on success and an error string on error:
...
+
+Recursive data protection can have a noticeable impact on memory usage when
+validating large data structures. If you are encountering issues with memory
+and you can guarantee that you do not have any loops in your data structure
+then deactivating the recursive data protection may help.
+
+B<Note: if you deactivate the recursive data protection and try to validate
+a data structure with a loop in it you'll send your program into an infinite
+loop - if in doubt leave it active!>
```
jhthorsen, line 1285: This description is too long. Also, this attribute *must* be marked as EXPERIMENTAL, and have a huge DISCLAIMER.

jlcooper: How  about if I change it to:

```
This attribute is EXPERIMENTAL and may change in a future release.

B<Disclaimer: Use at your own risk, if you have any doubt then don't use it>
```

jlcooper:
I've modified my debugging script to use some real life data (about 34MB) and logged the `.data` segment of the Perl process before validating the data and after validating the data.

```
With recursive_data_protection active:

                               data (kB)
  Before test data validated    87424
  After test data validated    752724
  Difference                   665300

With recursive_data_protection deactivated:

                               data (kB)
  Before test data validated    87400
  After test data validated    353868 
  Difference                   266468

Difference between active and inactive ~ 398 MB
```

lib/JSON/Validator.pm
jlcooper: Changed to a straight value.

lib/JSON/Validator.pm
jlcooper: Moved to a single line.

lib/JSON/Validator.pm
jlcooper: Switched to $boolean as suggested

lib/JSON/Validator.pm
jlcooper: Switched to the link as suggested

jlcooper:
I've added a unit test to test the recursive data protection logic.  It's a little more intrusive in the way that it tests the logic than I'd like, but we're having to test how it gets a result rather than the result it gets.

jhthorsen:
I'm not a fan of this at all, but I'm going to take the PR anyways, since the attribute is marked as EXPERIMENTAL. I have a gut feeling this can be fixed in a better way, but I don't know how.

Issue #222, opened 2020-09-30
Order in which errors are reported is undefined


davidcantrell-bb:
* JSON::Validator version: 4.02
* Perl version: 5.30.1
* Operating system: Mac OS Catalina

### Steps to reproduce the behavior
Validate some non-conforming data with multiple errors against a JSON schema

### Expected behavior
Errors are returned in a defined order

### Actual behavior
Errors are returned in an undefined order

That order *appears* to be consistently ASCIIbetical (sorted by the path to the offending data), both from a quick look at the code (although I didn't go through all of it) and from the results, so please could that be documented. This would make it easier to write and test code that does further processing on those errors.

jhthorsen:
This sounds like a good first time PR? I guess it could be fixed by adding `sort` in front of `keys` where "properties" and "patternProperties" are iterated in lib/JSON/Validator.pm, lib/JSON/Validator/Schema/Draft4.pm and lib/JSON/Validator/Schema/Draft6.pm.

The tricky part might be to write tests for this. Are you up for the task?

davidcantrell-bb:
Sure. Gimme a few days.

DrHyde:
Looks like a simpler way to make sure they're sorted would be to sort `@errors` based on the objects' `->path()`s before returning it from `validate()` would it not?

DrHyde:
Testing shouldn't be too difficult, just construct a schema and some known-bad data and randomise the order in which things appear in both. Do that a few times, and make sure that the errors are always returned in the same order.

jhthorsen:
@DrHyde: That could be a good idea, since the number of errors are probably much lower than the number of valid objects.

jhthorsen:
I'm closing this issue, since #223 will get merged.

PR #223, opened 2020-10-02
document and test that errors are spat out in a defined order

DrHyde:
The code change in lib/JSON/Validator.pm appears to be redundant at
the moment, as the tests passed without it, but I'm putting it there
anyway so that any future internals changes that affect ordering don't
suddenly start causing test failures.

### Summary
The list of errors spat out by `...->validate()` are now guaranteed to be in
a predictable order, which is documented.

### Motivation
Documenting the order of errors makes it easier to write tests for error conditions.

### References
https://github.com/mojolicious/json-validator/issues/222

### DIFF
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -187,7 +187,7 @@ sub validate {
   local $self->{seen}        = {};
   local $self->{temp_schema} = [];                            # make sure random-errors.t does not fail
   my @errors = $self->_validate($_[1], '', $schema);
-  return @errors;
+  return sort { $a->path() cmp $b->path() } @errors;
 }
 
 sub validate_json {
@@ -1356,7 +1356,9 @@ DEPRECATED.
   my @errors = $jv->validate($data, $schema);
 
 Validates C<$data> against a given JSON L</schema>. C<@errors> will
-contain validation error objects or be an empty list on success.
+contain validation error objects, in a predictable order (specifically,
+ASCIIbetically sorted by the error objects' C<path>) or be an empty
+list on success.
 
 See L</ERROR OBJECT> for details.
 
--- a/lib/JSON/Validator.pm
+++ b/lib/JSON/Validator.pm
@@ -186,8 +186,8 @@ sub validate {
   local $self->{schema}      = $self->_new_schema($schema);
   local $self->{seen}        = {};
   local $self->{temp_schema} = [];                            # make sure random-errors.t does not fail
-  my @errors = $self->_validate($_[1], '', $schema);
-  return sort { $a->path() cmp $b->path() } @errors;
+  my @errors = sort { $a->path cmp $b->path } $self->_validate($_[1], '', $schema);
+  return @errors;
 }
 
 sub validate_json {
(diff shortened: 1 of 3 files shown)


t/predictable-errors.t
```diff
@@ -0,0 +1,33 @@
+use strict;
+use warnings;
```
jhthorsen, line 2: You can “use Mojo::Base -strict” instead.

t/predictable-errors.t
```diff
@@ -0,0 +1,33 @@
...
+
+use JSON::Validator;
+use Math::Permute::List;
+use Test::More;
+
+my $jv = JSON::Validator->new();
+permute {
```
jhthorsen, line 9: I don’t see what value this adds. What’s wrong with a simple for-loop?

t/predictable-errors.t
```diff
@@ -0,0 +1,33 @@
...
+    $jv->schema(my $schema_text = join('',
+       '{ "type": "object", "properties": {',
+       join(', ', @_),
+       '}}'
+    ));
```
jhthorsen, line 14: You don’t have to give it a string. Just pass in the data structure instead.

jhthorsen (changes requested):
Thanks for the PR 👍

In addition to the inline comments: I don’t think the code is run through perltidy. Please do so, if not.

lib/JSON/Validator.pm
```diff
@@ -187,7 +187,7 @@ sub validate {
   local $self->{seen}        = {};
   local $self->{temp_schema} = [];                            # make sure random-errors.t does not fail
   my @errors = $self->_validate($_[1], '', $schema);
-  return @errors;
+  return sort { $a->path() cmp $b->path() } @errors;
```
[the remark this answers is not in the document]
perlpunk: Note that the behaviour of `sort` in scalar context is undefined.
If anyone is calling `validate` in scalar context to get the number of errors, with perl 5.30.3 for example it will return `undef`.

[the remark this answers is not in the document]
DrHyde: Ooh good catch @perlpunk, that's nasty

t/predictable-errors.t
DrHyde: I don't have control over the order of the properties in a perl data structure, I do it this way in case the order the JSON parser sees them makes a difference to the order in which they are processed.

t/predictable-errors.t
DrHyde: With a simple for loop I'd have to write out all the permutations by hand, this creates them on the fly.

jhthorsen: It's sort of irrelevant, since it is "randomized" by `keys()` inside JSON::Validator. I think it would be better to run this block 100 times with a fixed list.

Either way, I don't want the extra dependency just for this.

DrHyde: OK, done

jhthorsen:
Thanks for the PR. This is now part of version 4.05.

I did forget to squash the commits though. Please help me next time, by force pushing changes.

DrHyde:
Thanks

Issue #232, opened 2020-11-10
Given a valid JSON string that ends in a newline, the schema() method emits a warning


davidcantrell-bb:
* JSON::Validator version: 4.10
* Perl version: 5.30.1
* Operating system: Mac OS Catalina

### Steps to reproduce the behavior

```perl
JSON::Validator->new()->schema(q!{ "type": "object" }
!)
```

note the newline before the closing quote.

### Expected behavior
There should be no warning

### Actual behavior
A warning is emitted: `Unsuccessful stat on filename containing newline at /blahblah/JSON/Validator.pm line 317.`

### Suggested fix
Tightly scoped `no warnings 'newline';`, something like ...

```perl
my $loadable = do {
  no warnings 'newline';
  (blessed $source && ($source->can('scheme') || $source->isa('Mojo::File')))
  || -f $source
  || (!ref $source && $source =~ /^\w/);
};

abraxxa:
I just chased this error message for hours until I looked for open issues, thanks!

DrHyde:
There, have a trivial fix :-)

jhthorsen:
This is now fixed in #233.

PR #233, opened 2020-12-02
test and code-fix for issue 232

DrHyde:
### Summary
Tightly-scoped `no warnings 'newline'` in the chunk of code that tries to figure out whether something is a file or a string of JSON, to avoid unnecessarily spitting out a warning when it's a JSON string that ends in a newline.

### Motivation
The warning is wrong and confusing.

### References
https://github.com/mojolicious/json-validator/issues/232

### DIFF
--- a/t/newline-warnings.t
+++ b/t/newline-warnings.t
@@ -1,0 +1,15 @@
+use strict;
+use warnings;
+
+use Test::More;
+use JSON::Validator;
+
+my @warnings;
+$SIG{__WARN__} = sub { push @warnings, @_ };
+
+JSON::Validator->new()->schema(q!{ "type": "object" }
+!);
+
+ok(!@warnings, "no warning emitted when ->schema() method is passed a valid JSON schema ending in newline");
+
+done_testing;
--- a/t/newline-warnings.t
+++ b/t/newline-warnings.t
@@ -1,5 +1,4 @@
-use strict;
-use warnings;
+use Mojo::Base -strict;
 
 use Test::More;
 use JSON::Validator;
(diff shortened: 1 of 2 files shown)


abraxxa:
Isn't a chomp the easier fix?

DrHyde:
Errm, yes, if I'd remembered it existed :-)

t/newline-warnings.t
```diff
@@ -0,0 +1,15 @@
...
+use warnings;
+
+use Test::More;
+use JSON::Validator;
+
+my @warnings;
+$SIG{__WARN__} = sub { push @warnings, @_ };
+
+JSON::Validator->new()->schema(q!{ "type": "object" }
+!);
```
jhthorsen, line 11: I think adding "\n" is better. Right now it almost looks like a typo to me.

t/newline-warnings.t
```diff
@@ -0,0 +1,15 @@
+use strict;
+use warnings;
```
jhthorsen, line 2: Please use `use Mojo::Base -strict;` instead for consistency.

t/newline-warnings.t
```diff
@@ -0,0 +1,15 @@
...
+
+use Test::More;
+use JSON::Validator;
+
+my @warnings;
+$SIG{__WARN__} = sub { push @warnings, @_ };
+
+JSON::Validator->new()->schema(q!{ "type": "object" }
```
jhthorsen, line 10: Please remove parenthesis for consistency. (`new` vs `new()`)

jhthorsen (changes requested):
I'm not sure about this change. Can't we rather use a regexp to check if there's a newline?

jhthorsen:
Thanks!

PR #236, opened 2021-01-26
Make date-time regexp RFC3339 compatible

Skeeve:
RFC3339 states that the UTC offset needs to have 2 digit hour and minute offset.

Previous regexp allows for 1+ digits so "+1:1" and "-0123456789:9876543210" would be accepted.

### Summary
While using Openapi Plugin I noticed that it doesn't accept ISO timestamps. While this is fine, as Openapi spec states that the format be RFC3339, I noticed that the regexp checking the dates is too lax.

### Motivation
I see no advantage in accepting wrong time offsets like 123:567. 

### References
Discussion on IRC. Don't know how to link to it?

Issue #228, opened 2020-10-14
OpenAPIv2 and v3 schema classes should have introspection methods


mohawk2:
Further to needs expressed in #227, one of the grant-proposal authors writes, and I also find this valuable re introspection:

> Personally I'd like to see methods that formalise the existence of things that would otherwise be hash lookups, e.g. parameters, paths, models, maybe `get_model_by_name`, stuff like that

Reproduced here at request of @jhthorsen after it was deleted from #227.

jhthorsen:
This issue needs more details. What exactly is wanted? Please create a (partial) list of what you want to ask the schema of, and also a usecase for why that information is useful.

jhthorsen:
Here are some ideas:

* Get all request and response parameters by HTTP method, path and status code. This is required for validation #227
* Get all paths and HTTP methods.
* Get operation(s) by operationId or tags. This is useful for libraries such as OpenAPI::Client
* Get the base URL. This is useful for libraries such as OpenAPI::Client

jhthorsen:
Going to close this issue, since no more information has been provided. (I don't mind re-opening this though)

PR #237, opened 2021-01-27
add CLEAR method to J::V::Ref

mohawk2:
### Summary
JV 4.06 broke SQL::Translator::Parser::OpenAPI. This fixes it.

### Motivation
Breakage is bad.

### References
This would fix https://github.com/mohawk2/SQL-Translator-Parser-OpenAPI/issues/3

### DIFF
--- a/lib/JSON/Validator/Ref.pm
+++ b/lib/JSON/Validator/Ref.pm
@@ -31,6 +31,11 @@ sub FETCH {
   return undef;
 }
 
+sub CLEAR {
+  my ($self) = @_;
+  $self->[0] = {};
+}
+
 # Make it look like there is only one key in the hash
 sub FIRSTKEY { scalar keys %{$_[0][0]}; each %{$_[0][0]} }
 sub NEXTKEY  { each %{$_[0][0]} }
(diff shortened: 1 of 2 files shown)


jhthorsen (approved):
CLEAR has never been part of Ref.pm, so I'm curious why it started failing. Still seems good to have a complete API.

Something that you might be interested in is that JSON::Validator::OpenAPI::Mojolicious is going away in https://github.com/jhthorsen/mojolicious-plugin-openapi/pull/160

mohawk2:
Thanks for quick merge! May I ask if you have plans to release soon?

I expect previously JV::Ref was a hash-ref (I didn't review the changing commits), so the inherited methods worked.

Also, thanks for the heads-up on JVOAM going. I guess it's time to investigate using JV::OpenAPIv2 etc!

jhthorsen:
Soon-ish. I'm hoping to merge #236 first, so hopefully later today (Thursday).

Looking forward to having someone else use the new schema classes, and I wouldn't be surprised if you re-open #228 in the process.

mohawk2:
Yes, I also wouldn't be surprised!

PR #238, opened 2021-01-28
make date-time RFC3339 compatible

Skeeve:
RFC3339 states that the UTC offset needs to have 2 digit hour and minute offset.

Previous regexp allows for 1+ digits so "+1:1" and "-0123456789:9876543210" would be accepted.

### Summary
While using Openapi Plugin I noticed that it doesn't accept ISO timestamps. While this is fine, as Openapi spec states that the format be RFC3339, I noticed that the regexp checking the dates is too lax.

### Motivation
I see no advantage in accepting wrong time offsets like 123:567. 

### References
Discussion on IRC. Don't know how to link to it.

Spec: [OpenAPI](https://swagger.io/docs/specification/data-models/data-types/#string)
RFC: [RFC3339](https://datatracker.ietf.org/doc/rfc3339/?include_text=1)
See:
```
   time-hour       = 2DIGIT  ; 00-23
   time-minute     = 2DIGIT  ; 00-59
   time-numoffset  = ("+" / "-") time-hour ":" time-minute
```

### DIFF
--- a/lib/JSON/Validator/Formats.pm
+++ b/lib/JSON/Validator/Formats.pm
@@ -31,8 +31,10 @@ sub check_date {
 }
 
 sub check_date_time {
-  my @dt = $_[0] =~ m!^(\d{4})-(\d\d)-(\d\d)[T ](\d\d):(\d\d):(\d\d(?:\.\d+)?)(?:Z|([+-])(\d+):(\d+))?$!io;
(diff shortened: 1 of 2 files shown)


jhthorsen:
Oh. Didn't notice you made this PR until I released 4.13.

Duplicate of #236

PR #239, opened 2021-02-01
fixed SYNOPSIS extra semi-colon

akarelas:
### Summary
Removed a semi-colon from the SYNOPSIS that was causing a syntax error.


jhthorsen:
Thanks 👍

PR #251, opened 2021-06-23
Making J::V faster and easier to reason with

jhthorsen:
This PR can be tested with:

```
cpanm https://github.com/jhthorsen/json-validator/archive/v5_oh.tar.gz
```

This change will have some breaking changes, but hopefully it will be invisible for casual users.

Breaking changes:
* Got rid of JSON::Validator::Ref - 67e0b54a
* Removed support for "replace" in bundle() - a30b9b54
* Removed support for "allow_invalid_ref" in OpenAPIv2 - a30b9b54
* Moved schema_extract() into J::V::Schema - dd4611f8
* bundle() creates new definition names, and will keep the local document $ref's unchanged
* J::V->get() and J::V::Schema->get() behaves the same way
* Invalid $id or id in the schema will not confess, but result in schema validation errors

Enhancements:
* No need to sort as much, since @errors gets sorted - 6e94ce5c
* Split is_type() into smaller functions - 04162ed6
* Add J::V::Store->resolve() - 9ef7a5fc

Other internal changes:
* Going to call method on the schema object from J::V - a9793614, 427c8412
* Passing on $state to validation methods, instead of ($path, $schema) - 2fb1171f

Also added `t/benchmark.t` to track how the different changes affect what I'm trying to do: https://github.com/jhthorsen/json-validator/commits/v5_oh/t/benchmark.t. It shows significant improvements, but it's hard to say before all tests actually pass...

See also https://github.com/jhthorsen/mojolicious-plugin-openapi/pull/223


jhthorsen:
Current status:

v4.21:

Screenshot 2021-07-19 at 11-03-25 NYTProf benchmark t

This PR:

Screenshot 2021-07-19 at 11-00-45 NYTProf benchmark t

jhthorsen:
5.00 is now released 👍

PR #263, opened 2021-10-04
fix defaults for arrays with collectionFormat

elcamlost:
### Summary
Fixing defaults for arrays in query and path

### Motivation
Current implementation has a bug:
```yml
- name: elvishRingsOwners
  type: array
  items:
    type: string
  minItems: 1
  default: ["Gandalf", "Galadriel", "Cirdan"]
```
This schema cannot be validated if parameter is omitted.

### References
https://swagger.io/docs/specification/2-0/describing-parameters/#array-and-multi-value-parameters
there is an example of describing default value for array parameter

### DIFF
--- a/lib/JSON/Validator/Schema/OpenAPIv2.pm
+++ b/lib/JSON/Validator/Schema/OpenAPIv2.pm
@@ -333,6 +333,7 @@ sub _validate_request_or_response {
   my @errors;
   for my $param (@$parameters) {
     my $val = $self->_get_parameter_value($param, $get);
+    $self->_coerce_parameter_format($val, $param) if $direction eq 'request';
     $self->_coerce_default_value($val, $param) unless $val->{exists};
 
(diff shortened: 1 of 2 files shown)


jhthorsen (changes requested):
This PR looks incomplete. Did you forget to push some changes in that unit test?

I can’t accept the PR without a new test case.

elcamlost:
> This PR looks incomplete. Did you forget to push some changes in that unit test?
> 
> I can’t accept the PR without a new test case.


I changed t/openapiv2-collection-format.t test. Without changes in lib/JSON/Validator/Schema/OpenAPIv2.pm this test now fails. I can extract this changes into new test file, if it is what you're asking for.
But the new test is here ))

jhthorsen:
I tested the changes locally now and I see what you mean 👍 
Thanks for the PR!

PR #264, opened 2021-11-11
fix joi required() with nested objects

TimStallard:
### Summary
This PR fixes Joi's required() method when used with nested objects.

### Motivation
Currently the `required` attribute is used for both the boolean flag of whether a property is required, and also to store the array of required properties. When used with nested objects, this either ends up with properties incorrectly being marked as required, or dying with `Can't use string ("1") as an ARRAY ref while "strict refs" in use`.

This PR moves the array of required properties into a separate attribute.


jhthorsen (approved):
Looks great! Thank you 👍

Issue #261, opened 2021-09-20
Wide character in subroutine entry


xurc:
* JSON::Validator version: 4.24
* Perl version: 5.28.1
* Operating system: Debian GNU/Linux 10

### Steps to reproduce the behavior
Add an array with string unique items to the specification ("uniqueItems": true)
```
           "test": {
             "type": "array",
             "uniqueItems": true,
...
           }
```
and validate non-ASCII values
```
          "test" => [
            "\x{422}\x{435}\x{441}\x{442}1",
            "Test1"
          ],
```

### Expected behavior
No errors

### Actual behavior
`Wide character in subroutine entry at /home/crux/workspace/qq/core/lib/JSON/Validator/Util.pm line 27.`

Apparently the error occurs because the Mojo::Util::md5_sum is expecting bytes here:

```
package JSON::Validator::Util;
...
sub data_checksum {
  return Mojo::Util::md5_sum(ref $_[0] ? $serializer->($_[0]) : defined $_[0] ? qq('$_[0]') : 'undef');
}
```
While stuck in a local dirty hotfix:
```
sub data_checksum {
  my $t = $_[0];
  ref $t ? 1 : defined $t && utf8::is_utf8($t) ? utf8::encode($t) : 1;
  return Mojo::Util::md5_sum(ref $_[0] ? $serializer->($_[0]) : defined $_[0] ? qq('$t') : 'undef');
}
```

jhthorsen:
I’m pretty sure you have to make sure you have bytes everywhere, meaning you have to encode your data before passing it to the validator.

xurc:
> 
> 
> I’m pretty sure you have to make sure you have bytes everywhere, meaning you have to encode your data before passing it to the validator.

This is counterintuitive, since UTF comes from the database, and UTF leaves in response. Applications generally work with characters, not bytes.

In this case, bytes are required only due to the fact that inside the validator to check uniqueness, md5 is used, which needs bytes, it is not logical to convert all data due to the internal implementation in the library.

jhthorsen:
I'm going to close this issue, since there haven't been any updates for a year.

PR #268, opened 2022-10-25
Allow unicode strings in the JSON schema

iamb:
### Summary
Allow JSON::Validator to validate schema with unicode strings.

### Motivation

I ran into this problem with a 3rd party json schema over which I have no control.

### References
This is the same issue described in #261 although I ran into it with an enum rather than uniqueItems.

The user of JSON::Validator does not have control over bytes vs. strings as given to data_checksum. When a schema comes from JSON, strings are decoded into strings of perl characters. The YAML modules decode to perl strings as well.

I had a similar fix to the one mentioned in that issue with a couple exceptions:

I did not copy the argument unless necessary because I wasn't sure if that was intentionally avoided in the original code.

I encode unconditionally the stringified scalar which is fed into the digest. This is intentional to avoid incorrectly accepted doubly-encoded UTF-8 as valid. That is demonstrated in the included tests.

I ran t/benchmark.t with n=200, n=500, and n=1000 against the previous code and this updated code and was not able to get any differences beyond noise.

### DIFF
--- a/lib/JSON/Validator/Util.pm
+++ b/lib/JSON/Validator/Util.pm
@@ -24,7 +24,7 @@ sub E { JSON::Validator::Error->new(@_) }
 my $serializer = SEREAL_SUPPORT ? \&_sereal_encode : \&_yaml_dump;
 
 sub data_checksum {
-  return Mojo::Util::md5_sum(ref $_[0] ? $serializer->($_[0]) : defined $_[0] ? qq('$_[0]') : 'undef');
+  return Mojo::Util::md5_sum(ref $_[0] ? $serializer->($_[0]) : defined $_[0] ? do { utf8::encode(my $x = shift); qq('$x') } : 'undef');
 }
 
 sub data_section {
--- a/t/spec/with-unicode-multibyte.json
+++ b/t/spec/with-unicode-multibyte.json
@@ -2,7 +2,7 @@
   "type": "object",
   "properties": {
     "foo": {
-      "enum": ["foo–bar"]
+      "enum": ["foo♫bar"]
     }
   }
 }
--- a/t/spec/with-unicode-multibyte.json
+++ b/t/spec/with-unicode-multibyte.json
@@ -1,6 +1,9 @@
 {
   "type": "object",
   "properties": {
+    "bar": {
+      "enum": ["replacement�char"]
+    },
     "foo": {
       "enum": ["foo♫bar"]
     }
--- a/t/spec/with-unicode-multibyte.json
+++ b/t/spec/with-unicode-multibyte.json
@@ -1,0 +1,8 @@
+{
+  "type": "object",
+  "properties": {
+    "foo": {
+      "enum": ["foo–bar"]
+    }
+  }
+}
--- a/t/unicode-multibyte.t
+++ b/t/unicode-multibyte.t
@@ -9,15 +9,23 @@ binmode($builder->output, ':encoding(UTF-8)');
 binmode($builder->failure_output, ':encoding(UTF-8)');
 binmode($builder->todo_output, ':encoding(UTF-8)');
 
-my $perl_utf8_str = "foo\x{266b}bar";
-my $encoded_bytes = encode('UTF-8', $perl_utf8_str);
 my $json_file = path(__FILE__)->dirname->child('spec')->child('with-unicode-multibyte.json');
 my $yaml_file = path(__FILE__)->dirname->child('spec')->child('with-unicode-multibyte.yml');
+
+my $perl_utf8_str = "foo\x{266b}bar";
+my $encoded_bytes = encode('UTF-8', $perl_utf8_str);
+my $with_replacement_char = "replacement\x{fffd}char";
+my $invalid_utf8_1 = "replacement\x{d800}char";
+my $invalid_utf8_2 = "replacement\x{d8f0}char";
 
 validate_ok {foo => $perl_utf8_str}, $json_file;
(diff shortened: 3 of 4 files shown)


lib/JSON/Validator/Util.pm
```diff
@@ -24,7 +24,7 @@ sub E { JSON::Validator::Error->new(@_) }
 my $serializer = SEREAL_SUPPORT ? \&_sereal_encode : \&_yaml_dump;
 
 sub data_checksum {
-  return Mojo::Util::md5_sum(ref $_[0] ? $serializer->($_[0]) : defined $_[0] ? qq('$_[0]') : 'undef');
+  return Mojo::Util::md5_sum(ref $_[0] ? $serializer->($_[0]) : defined $_[0] ? do { utf8::encode(my $x = shift); qq('$x') } : 'undef');
```
jhthorsen, line 27: Could you use encode from Mojo::Util instead? I also wonder what would happen if we run everything through $serializer...

t/spec/with-unicode-multibyte.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "object",
+  "properties": {
+    "foo": {
+      "enum": ["foo–bar"]
```
jhthorsen, line 5: Could you use a heart or something else that is easier to read than "–" ? Example: "json-validater ♥️ utf8"

t/unicode-multibyte.t
```diff
@@ -0,0 +1,23 @@
+use lib '.';
+use t::Helper;
+use Mojo::File 'path';
+use Mojo::Util 'encode';
```
jhthorsen, line 4: Can you use `qw(...)` instead of single quotes for consistency?

t/unicode-multibyte.t
```diff
@@ -0,0 +1,23 @@
...
+use Test::More;
+
+my $builder = Test::More->builder;
+binmode($builder->output, ':encoding(UTF-8)');
+binmode($builder->failure_output, ':encoding(UTF-8)');
+binmode($builder->todo_output, ':encoding(UTF-8)');
```
jhthorsen, line 10: I'm surprised that you have to do this, since I've never done that in any other unicode related tests. Can you explain?

jhthorsen (changes requested):
I would be surprised if this PR doesn't get merged one way or another, but I would like to see some changes before that happens.

lib/JSON/Validator/Util.pm
iamb: I could, but chose specifically to use utf8::encode specifically. It shouldn't be possible for invalid codepoints (e.g., `\x{d800}`) to appear, but if they somehow did Mojo::Util (or Encode directly) would insert replacement characters. Different inputs (with invalid codepoints in the same place) would hash the same and compare the same. With utf8::encode, the perl representation is directly put into codepoints in the resulting string even if they are invalid. This way, different perl strings will always hash different.

If using Mojo::Util::encode

t/spec/with-unicode-multibyte.json
iamb: Sure, will change. Happened to be the character I ran into.

t/unicode-multibyte.t
iamb: Will change

t/unicode-multibyte.t
iamb: This is just to avoid warnings when TAP is printing...for example:

````
t/unicode-multibyte.t .. 1/? Wide character in print at /opt/perl/perls/5.36.0/lib/5.36.0/Test2/Formatter/TAP.pm line 156.
Wide character in print at /opt/perl/perls/5.36.0/lib/5.36.0/Test2/Formatter/TAP.pm line 156.
t/unicode-multibyte.t .. ok
All tests successful.
Files=1, Tests=4,  0 wallclock secs ( 0.02 usr  0.01 sys +  0.23 cusr  0.02 csys =  0.28 CPU)
Result: PASS
0
````

I'm not sure it's the best way either, but I don't usually deal with output unicode from tests. I can remove it if you prefer.

lib/JSON/Validator/Util.pm
jhthorsen: If that's the case then you have to add a(nother) test for that, or else I can pretty much guarantee you that I will change this later on and introduce a regression.

t/unicode-multibyte.t
jhthorsen: Sounds like validate_ok() should be fixed instead? If you think validate_ok() should be changed then please remove these three lines. If you want/can then please fix validate_ok() 😄

lib/JSON/Validator/Util.pm
iamb: gotcha, test added

t/unicode-multibyte.t
iamb: I'm pretty sure validate_ok() is correct here, the messages it's outputting are perl strings and it's up to the tap formatter to deal with output encoding. I need to research how to run tests better with expected unicode output (checking mojo for example, test names explicitly use "\uNNNN" rather than the characters)...I think that's on me. For now, I've removed from the tests.

lib/JSON/Validator/Util.pm
iamb: FWIW, I ran the benchmark test with n=1000 running this vs. just using serializer for everything. Using serializer was something like 35% slower in validating (just judging by repored time from the test).

jhthorsen: Thanks for checking! I was expecting something like that, but I wasn't sure.

jhthorsen (approved):

PR #275, opened 2024-08-08
make JSON::Validator::Util::is_bool return true when passed perl v5.36+ builtin booleans

akarelas:
### Summary

This PR allows newer Perls' (v5.36+) builtin core booleans to be validated as booleans. It's quite a necessary change to make, as when you include the line `use v5.40` in your code file, you may not replace the meaning of `true` and `false` from perl's core booleans just by writing `use Mojo::JSON 'true', 'false';`.  Therefore, it seems, core booleans need to be considered valid by all CPAN modules.

I did it by modifying slightly the `JSON::Validator::Util::is_bool` function, inspired by how `JSON::PP` recently achieved the same effect.

### Motivation

My motivation was to resolve this bug of Mojolicious::Plugin::OpenAPI: https://github.com/jhthorsen/mojolicious-plugin-openapi/issues/252

### Tried my solution

I tested my modifications on a test Mojolicious website that uses Mojolicious::Plugin::OpenAPI, using both perl v5.34 (that doesn't have the builtin booleans) and perl v5.40 (that does), and the site works in both these cases.

I forgot to run `prove -vl` on it, or write tests with core booleans that will run only on perl v5.40. I will do so later today or tomorrow.

### References

* How JSON::PP did it: https://metacpan.org/dist/JSON-PP/source/lib/JSON/PP.pm#L208-216
* Bug of Mojolicious::Plugin::OpenAPI: https://github.com/jhthorsen/mojolicious-plugin-openapi/issues/252
* On Monday, June 10th, 2024 you expressed interest in a PR to fix this, on the perl-openapi channel on IRC

### DIFF
--- a/lib/JSON/Validator/Util.pm
+++ b/lib/JSON/Validator/Util.pm
@@ -13,6 +13,7 @@ use Mojo::Util;
 use Scalar::Util 'blessed';
 
 use constant SEREAL_SUPPORT => !$ENV{JSON_VALIDATOR_NO_SEREAL} && eval 'use Sereal::Encoder 4.00;1';
+use constant CORE_BOOL      => defined &builtin::is_bool;
 
 our @EXPORT_OK = (
   qw(E data_checksum data_section data_type is_bool is_num is_type),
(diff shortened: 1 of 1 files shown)


jhthorsen:
I can't take the change, since too many tests fail.

akarelas:
We need to find out why those tests fail. Would they fail on any PR (even adding a blank line in a file)? If so, it's really a problem not to be attributed to me. If not, then I need to look at it.

jhthorsen:
Going to take your PR! Looks like you were spot on in #279. 🤞

PR #284, opened 2025-04-12
Dirty fix

dariuszpiwowarski:
(https://github.com/jhthorsen/mojolicious-plugin-openapi/issues/261)

### DIFF
--- a/lib/JSON/Validator/Schema/OpenAPIv3.pm
+++ b/lib/JSON/Validator/Schema/OpenAPIv3.pm
@@ -289,12 +289,13 @@ sub _validate_body {
     return E "/$param->{name}", [qw(object required)];
   }
   if ($val->{exists}) {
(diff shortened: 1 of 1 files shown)


lib/JSON/Validator/Schema/OpenAPIv3.pm
```diff
@@ -289,12 +289,13 @@ sub _validate_body {
     return E "/$param->{name}", [qw(object required)];
   }
   if ($val->{exists}) {
+    my $negotiated_content_type = negotiate_content_type($param->{accepts}, $val->{content_type});
```
archey347, line 292: It might be worth adding on something like:

```suggestion
    my $negotiated_content_type = negotiate_content_type($param->{accepts}, $val->{content_type}) || $param->{accepts}[0];
```

so validation isn't also bypassed if the content type is an empty string

PR #288, opened 2025-11-19
Add file placeholder

archey347:
### Summary
Allow using `Mojo::Upload` that can be used as a placeholder for files
- Is allowed when `format: binary` is defined on strings: https://swagger.io/docs/specification/v3_0/describing-request-body/file-upload/)

### Motivation
See https://github.com/jhthorsen/mojolicious-plugin-openapi/pull/266

Allows for files to be validated against without having to make a clone of the file.

### References
Is depended on by https://github.com/jhthorsen/mojolicious-plugin-openapi/pull/266

### DIFF
--- a/lib/JSON/Validator/Util.pm
+++ b/lib/JSON/Validator/Util.pm
@@ -59,6 +59,7 @@ sub data_type {
   my $ref     = ref $_[0];
   my $blessed = blessed $_[0];
   return 'object'  if $ref eq 'HASH';
+  return 'file'    if $ref eq 'Mojo::Upload';
   return lc $ref   if $ref and !$blessed;
   return 'null'    if !defined $_[0];
   return 'boolean' if $blessed and ("$_[0]" eq "1" or !"$_[0]");
--- a/lib/JSON/Validator/Schema/OpenAPIv3.pm
+++ b/lib/JSON/Validator/Schema/OpenAPIv3.pm
@@ -407,8 +407,19 @@ sub _validate_type_object_response {
 }
 
 sub _validate_type_string {
-  my $self = shift;
-  return $_[1]->{schema}{nullable} && !defined $_[0] ? () : $self->SUPER::_validate_type_string(@_);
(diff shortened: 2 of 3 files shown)


lib/JSON/Validator/Util.pm
```diff
@@ -62,6 +62,7 @@ sub data_type {
   return lc $ref   if $ref and !$blessed;
   return 'null'    if !defined $_[0];
   return 'boolean' if $blessed and ("$_[0]" eq "1" or !"$_[0]");
+  return 'file' if $ref eq 'JSON::Validator::FilePlaceholder';
```
archey347: nitpick: misaligned if statement

lib/JSON/Validator/FilePlaceholder.pm
```diff
@@ -0,0 +1,85 @@
...
+sub new {
+  my ($class, $options) = @_;
+
+  my $file = {
+    filename  => $options->{filename},
+    size      => $options->{size},
+  };
+
+  return bless $file, $class;
+}
```
jhthorsen: Why can't you just use new() from Mojo::base?

lib/JSON/Validator/FilePlaceholder.pm
```diff
@@ -0,0 +1,85 @@
...
+
+sub filename {
+  return $_[0]->{file}{filename};
+}
+
+sub size {
+  return $_[0]->{file}{size};
+}
```
jhthorsen: I don't understand where the "file" key comes from. Are there some missing tests?

jhthorsen (changes requested):
I would prefer if you could use https://metacpan.org/pod/Mojo::Upload instead of creating a new class.

lib/JSON/Validator/FilePlaceholder.pm
archey347: I can't see any reason why it needs it either, likely a mistake on my part.

Issue #272, opened 2024-02-13
heisenbug with openapi 3.1.* schema


vividsnow:
* JSON::Validator version: 5.14
* Perl version: 5.36, 5.38 (non-system)
* Operating system: debian stable/unstable

### Steps to reproduce the behavior
Given script:
```perl
use strict;
use JSON::Validator::Schema::OpenAPIv3;

my $yaml = <<~\in;
openapi: 3.1.0

info:
  version: 1.0
  title: whatever

components:
  parameters:
    Id:
      in: path
      name: id
      schema:
        $ref: '#/components/schemas/IdType'
      required: true

  schemas:
    IdType:
      type: integer
in

my ($ok, $nok) = (0, 0);
for (1..100) {
    my $js = JSON::Validator::Schema::OpenAPIv3->new(
        $yaml,
        specification => 'https://spec.openapis.org/oas/3.1/schema/2021-05-20'
    );
    $ok++, next unless $js->is_invalid;
    $nok++;
    print "$_\n" for @{$js->errors};
}
print "ok: $ok, nok: $nok\n";
```

running as `perl -E 'system q(perl json-validator.pl) for 1..10'`

### Expected behavior
Expecting having valid schema every run:
```bash
...
ok: 100, nok: 0
...
```

### Actual behavior
Getting random fails:
```bash
ok: 100, nok: 0
/components/parameters/Id/$ref: Missing property.
...
ok: 95, nok: 5
/components/parameters/Id/$ref: Missing property.
...
ok: 97, nok: 3
ok: 100, nok: 0
ok: 100, nok: 0
/components/parameters/Id/$ref: Missing property.
...
ok: 97, nok: 3
/components/parameters/Id/$ref: Missing property.
...
ok: 95, nok: 5
/components/parameters/Id/$ref: Missing property.
...
ok: 86, nok: 14
/components/parameters/Id/$ref: Missing property.
...
ok: 97, nok: 3
/components/parameters/Id/$ref: Missing property.
...
ok: 96, nok: 4
```

jhthorsen:
Fixed in https://github.com/jhthorsen/json-validator/pull/294

PR #294, opened 2026-06-06
Fix nondeterministic false errors from the recursive_data_protection cache

HEM42:
### Summary

Keep the cached `$schema` and `$data` referenced for the duration of the validation by storing
them in the cache value (`[\@errors, $schema, $data]`), so their addresses cannot be reused by
later transient schemas in the same run. This preserves the existing memoization and
cyclic-data-recursion semantics - `t/recursive_data_protection.t` and `t/validate-recursive.t`
still pass - and only removes the address-reuse collision.

### Root cause

`recursive_data_protection`'s `seen` cache in `JSON::Validator::Schema::_validate` keys on
stringified memory addresses (`"$schema:$data"`) and keeps entries for the whole validation. When
a schema follows many `$ref`s (the OpenAPI 3.1 meta-schema does), `_state` builds transient merged
schema hashrefs; these are freed mid-validation and their addresses get reused, so a later cache
key collides with a freed entry and returns a stale error list - here the discarded
`if: { required: ["$ref"] }` branch error. Because it depends on address reuse, the failure is
intermittent and unrelated to `PERL_PERTURB_KEYS`/hash ordering.

### References

#272

### DIFF
--- a/lib/JSON/Validator/Schema.pm
+++ b/lib/JSON/Validator/Schema.pm
@@ -372,8 +372,13 @@ sub _validate {
   my @errors;
   if ($self->recursive_data_protection and 2 == grep { ref $_ && !is_bool($_) } $data, $schema) {
     my $seen_addr = "$schema:$data";
-    return @{$self->{seen}{$seen_addr}} if $self->{seen}{$seen_addr};    # Avoid recursion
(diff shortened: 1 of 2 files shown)


jhthorsen:
Tests are passing, so it's a merge! 👍 Thanks

PR #299, opened 2026-06-09
Fix OpenAPIv3 string coercion regression

archey347:
### Summary
The override of _validate_type_string introduced in a8fe5fa unpacked @_ into lexical copies and passed those to SUPER.

However, the call to SUPER mutates the values in @_, which I'd incorrectly assumed it didn't.

Credit to @NGEfreak for spotting and the actual fix.

### References
Fixes https://github.com/jhthorsen/json-validator/issues/298

### DIFF
--- a/lib/JSON/Validator/Schema/OpenAPIv3.pm
+++ b/lib/JSON/Validator/Schema/OpenAPIv3.pm
@@ -407,7 +407,8 @@ sub _validate_type_object_response {
 }
 
 sub _validate_type_string {
-  my ($self, $val, $state) = @_;
+  my $self = shift;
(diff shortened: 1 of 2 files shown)


lib/JSON/Validator/Schema/OpenAPIv3.pm
```diff
@@ -407,7 +407,8 @@ sub _validate_type_object_response {
 }
 
 sub _validate_type_string {
-  my ($self, $val, $state) = @_;
+  my $self = shift;
+  my ($val, $state) = @_;
```
jhthorsen, line 411: This looks very strange, but i'll take the PR 👍

